PR #19, opened 2020-04-10
Update guidelines

JensWehner:
- Removed the developer.pdf
- Translated most of it into the DEVELOPERS_GUIDE.md
- removed the file header.txt and put it into the DEVELOPERS_GUIDE.md
- added a section about formatting
- added ``.clang-format`` file.


DEVELOPERS_GUIDE.md
```diff
@@ -0,0 +1,203 @@
...
+- When using a template, use typename T and NOT class T.
+- Avoid using ``typename`` as a return specifier.
+- Try to avoid ``typedef`` for type definitions. Use the ``using``-keyword if either the type is likely to change, or the typename is very long. This should be done in the class declaration and not globally.
+- Avoid returning by final function argument (for example ``void myFun (int a, int b, Matrix mat)`` for computing mat).
+
+
+
+## About assertions
+
+The intended use of assertions in general and ``logger.assert`` for hpGEM is to check blatant assumptions on the arguments provided to the
```
Laar, line 113: We have changed `logger.assert` and now have two options, `logger.assert_debug` and `logger.assert_always`. The first one is compiled out in release mode. So I think we might need to rewrite this section a bit. My general feel is that
 
 - assert_always is where removing the check is not expected to make any (significant) speed gain or things that must always be checked. Thus, usage is restricted to checks that are cheap in code that runs infrequently. Furthermore, it is used in unittests.
 - assert_debug, used everywhere else in liberal fashion.

DEVELOPERS_GUIDE.md
```diff
@@ -0,0 +1,203 @@
...
+### Naming Conventions
+
+- File names must be of the form ``FileName``.
+- Directory names must be of the form ``DirectoryName``.
+- Names representing types must be in mixed case starting with upper case, i.e. ``TypeName``.
+- Variable names must be in mixed case starting with lower case, i.e. ``variableName``.
+- Named constants (including enumeration values) must be all uppercase using underscore to separate words.
+- Names representing methods or functions must be verbs and written in mixed case starting with lower case, i.e. ``functionName``.
+- Names representing Classes must be of the form ``ClassName``.
+- Names representing template types should be a single uppercase letter.
```
Laar, line 53: This is quite often violated, with `DIM` as template parameter for the dimension being the canonical example. Maybe we should reformulate this one.

DEVELOPERS_GUIDE.md
```diff
@@ -0,0 +1,203 @@
...
+
+### Files
+
+- C++ header files should have the extension ``.h``.
+- Source files should have the extension ``.cpp``.
+- Application files should have the extension ``.cpp``.
+- Implementation files of templated classes should have the extension ``_Impl.h`` and be included at the end of the header file.
+- Large source files can be split in multiple files by using the same name, then an ``_`` and a description of its usage.
+- A class should be declared in a header file and defined in a source file where the name of the files match the name of the class.
+- All definitions should reside in source files.
```
Laar, line 72: This is violated so many times in the current state of the repository. Probably to allow the compiler to inline functions. I do not know a good alternative guideline at the moment.

DEVELOPERS_GUIDE.md
```diff
@@ -0,0 +1,203 @@
...
+- Special characters like TAB and page break must be avoided.
```
Laar, line 73:
```suggestion
- Use spaces for indenting, the use of tab should be avoided.
```
I think this may be clearer

DEVELOPERS_GUIDE.md
```diff
@@ -0,0 +1,203 @@
...
+- Multiple variables should not be declared on the same line; except for the primitive data types e.g. ``int``.
+- Use of global variables should be minimized.
+- Class variables should never be declared public.
+- C++ pointers and references should have their reference symbol next to the name rather than to the type, i.e. ``Element *element``.
+- Implicit test for 0 should not be used other than for boolean variables.
+- Variables should be declared in the smallest scope possible.
+- Only loop control statements must be included in the ``for()`` construction.
+- Range based for-loops are preferred over other types of for-loops. For example, a for-loop over vector vec can look like ``for ( double d: vec) { }``.
+- Loop variables should be initialized immediately before the loop.
+- Do-while loops can be avoided.
```
Laar, line 94:
```suggestion
- Do-while loops should be avoided.
```
The previous one is a statement not an instruction

DEVELOPERS_GUIDE.md
```diff
@@ -0,0 +1,203 @@
...
+- The use of ``break`` and ``continue`` in loops should be avoided, go for readability.
+- The form ``while (true)`` should be used for infinite loops.
+- In cases of an if-statement, the nominal case should be put in the if-part and the exception in the else-part.
+- Floating point constants should always be written with decimal point and at least one decimal. (i.e. ``x = 0.0``).
+- Use ``nullptr`` instead of ``NULL`` or ``0``.
+- Try to avoid smart pointers. If a smart pointer is necessary, use unique_ptr or shared_ptr.
+- Use the type ``std::size_t`` for non-negative integer variables.
+- Encapsulate global variables and constants, enumerated types, and type definitions in a class or namespace.
+- The ``auto``-keyword for type declaration is only allowed if it is instantly clear what the type is.
+- When using a template, use typename T and NOT class T.
```
Laar, line 104:
```suggestion
- When using a template, use ``typename T`` and NOT ``class T``.
```

Just formatting

README.md
```diff
@@ -52,12 +52,16 @@ describes a mesh of 7x8 triangular elements on the boundary of a cylinder with h
...
-For questions, comments, requests and bug reports, send an e-mail to support@hpgem.org
+For questions, comments, requests and bug reports, send an e-mail to support@hpgem.org or open an issue [here](https://github.com/hpgem/hpgem/issues).
  
 ## More information
  
 More information is available on the website: http://www.hpgem.org
 
+## Contributing
+
+Contributions are very welcome, please open an issue [here](https://github.com/hpgem/hpgem/issues) before submitting a Pull-Request, for larger changes. So that they can be discussed. You find more information in the [Developers Guide](https://github.com/JensWehner/hpgem/blob/update_guidelines/DEVELOPERS_GUIDE.md).
```
Laar, line 63: Does the link needs to be changed afterwards?

Laar:
I found the following documentation on the mesh format in the previous wiki with the development guidelines. It may be included here, but I think it is better to split it of.


## Mesh files

Starting from revision 916 mesh generation has been extracted to a preprocessor. This is required because we don't have enough RAM available to store the full mesh on every processor for large parallel simulations. The preprocessor avoids this problem by running sequentially and it does not need all the extra memory to store FEM solutions. The output of the preprocessor is such that each process in the large simulation only has to read and store the elements that it requires. Documentation about acceptable inputs for the preprocessor can be found in the README file in the project root directory. The rest of this section will describe the mesh files. For examples you can look in tests/files, which contains input data in files containing the word source and corresponding output in similarly named files containing the word mesh. The first line of the mesh files always is equal to mesh 1. The next line contains 3 integers denoting the total number of nodes n, the total number of elements e and the dimension d of the mesh. In files describing meshes of dimension 2 or higher this is followed by the number of faces f in the mesh and for files describing 3-dimensional meshes also the number of edges e in the mesh.
The third line contains an integer denoting the number of partitions p, followed by p integers denoting in order the number of nodes that will be attributed to that partition. Note that the sum of these local amounts of nodes will in general be larger than the total number of nodes, because at the boundaries of partitions nodes need to appear in multiple partitions. After this there will be two lines per node. The first of these contains an integer denoting the number of partitions that should read this node p_n followed by p_n integers denoting those partitions. The other line contains an integer denoting the number of coordinates this node is associated with c, followed by c d-tuples of floating point numbers. Generated files will have these numbers in hexadecimal format to prevent rounding errors, but if you need to tweak one of these files a little, any floating point format will work. Nodes are linked to more than one coordinate if they are positioned on the 'boundary' of a periodic domain.
After the information about the nodes there is one line per element containing an integer representing the number of nodes for that elements n_e followed by n_e pairs of integers representing the global index of the nodes and the internal index of the coordinate that should be active in this element respectively. It contains all information for one node before continuing to the next node. After this it contains an integer representing the partition that is responsible for computations on this element, followed by an integer representing the number of partitions this element appears in as a shadow element s, followed by s integers denoting those shadow partitions. The global index of a node according to this file is equal to the number of nodes that precede it in the listing of nodes.
For files of dimension 2 or higher the remainder of the file contains entities that are not elements or nodes, ordered such that all entities of a higher dimension are listed before entities with a lower dimension. There is one line per entity. It begins with an integer denoting the number of elements n_f adjacent to that entity. After this there are n_f pairs of integers denoting the global index of the element and the local index of the entity inside the element respectively. It contains all information for one element before continuing the the next element. After this it contains an integer representing the number of partitions p_e that this entity should appear in. The line end with p_e integers denoting those partitions. Not that a partition is only guaranteed to contain all adjacent elements if it is responsible for computations on at least one adjacent element.

README.md
JensWehner: yes. I will do that afterwards. Links to not yet created files are a bit awkward.

DEVELOPERS_GUIDE.md
JensWehner: I think the best way is to not put it in the class declaration, but underneath in the same file. I agree that putting it in a separate header is not beneficial.

JensWehner:
closed because I want to get rid of my fork. See https://github.com/hpgem/hpgem/pull/28

PR #21, opened 2020-04-10
removed unneeded files

JensWehner:
In the main source directory where 4 files, which according to `grep` are used nowhere. This deletes them. 

I have not tested it further atm.


Laar:
Tests run

PR #31, opened 2020-05-26
Remove old mesh generation/reading from kernel

### LINKED ISSUE #26: Deprecation warnings for CentaurMeshReaderUnitTest.cpp with Qhull
/home/jens/Programs/hpgem/tests/unit/Base/CentaurMeshReaderUnitTest.cpp:177:50: warning: ‘void Base::MeshManipulator<DIM>::readCentaurMesh(const string&) [with long unsigned int DIM = 2; std::__cxx11::string = std::__cxx11::basic_string<char>]’ is deprecated: This will not work optimally for parallel computations use the preprocessor instead [-Wdeprecated-declarations]
     myTwoDDemoMesh.readCentaurMesh(filename.str());

Laar:
The old (before the preprocessor) ways of reading and generating meshes
were accidentally readded in 12198e60. They were kept due to some bugs
causing issues with the same meshes from the preprocessor. As these bugs
all seem to have been fixed, we now again remove the generation of
square meshes in the kernel, just as the support for reading centaur
meshes directly.

Currently in WIP status as it is untested with a real petsc.

This also resolves #26.


Laar:
Now also tested with real petsc & slepc.

PR #34, opened 2020-05-28
Clang tidy

JensWehner:
Updated the code with the following clang-tidy checks:

modernize-loop-convert
modernize-make-shared
modernize-use-using
modernize-make-unique
readability-else-after-return
readability-const-return-type
readability-simplify-boolean-expr
readability-redundant-smartptr-get
modernize-use-nullptr
modernize-use-override
modernize-deprecated-headers
modernize-use-equals-default
modernize-use-equals-delete

merge after: https://github.com/hpgem/hpgem/pull/33


Laar:
Might not be the biggest fan of  readability-else-after-return, but looks good.

PR #38, opened 2020-05-29
added macos CI

JensWehner:
- added macos CI
- fixed some spelling errors
- expanded codecov build

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -1,0 +1,28 @@
+name: MacOs
+on:
+  push:
+    branches: [ master ]
+  pull_request:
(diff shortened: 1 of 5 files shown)


.github/workflows/macos.yml
```diff
@@ -0,0 +1,28 @@
...
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: macos-latest
+    strategy:
+      matrix:
+        mpi: [ ON, OFF ] 
+        compiler: [gcc]
```
Laar, line 15: For me it would seem more logical to test with clang as that is the default compiler on Mac.

PR #39, opened 2020-06-02
Refinement fixes

Laar:
Some simple fixes to make the mesh refinement function compile again. This was broken by the introduction of explicit ownership in b4d28c9 (and some later commits), but the compiler did not detect it.


JensWehner:
@Laar we should not make the Qhulll builds required.

Laar:
Indeed we should, but I can not find how to do it at the moment. As it seems that I can switch the build configurations (MPI ON/OFF, clang/gcc, debug/release) but not the prefix (Basic, Petsc, QHull).

PR #40, opened 2020-06-08
Unify dgmax eigenvalue

Laar:
Make a shared interface between `DGMaxEigenvalue` and `DivDGMaxEigenvalue` so that code can be written that works independently of the algorithm used to solve the eigenvalue problem.

The changes roughly include:
 - Renaming EigenValue -> Eigenvalue (as eigenvalue is a single word in English)
 - Introduce a shared interface `AbstractEigenvalueSolver`
 - Move all configurations parameters of both solvers to the constructor (to satisfy the interface)


JensWehner:
formatting is not correct yet.

applications/DG-Max/Algorithms/DivDGMaxEigenvalue.cpp
```diff
@@ -36,25 +36,25 @@ ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 
-#include "DivDGMaxEigenValue.h"
+#include "DivDGMaxEigenvalue.h"
 
-#include "../DGMaxLogger.h"
+#include <valarray>
```
JensWehner, line 41: why valarray ?

applications/DG-Max/Algorithms/DivDGMaxEigenvalue.cpp
```diff
@@ -357,12 +357,12 @@ typename DivDGMaxEigenValue<DIM>::Result DivDGMaxEigenValue<DIM>::solve(
     error = EPSDestroy(&eigenSolver);
     CHKERRABORT(PETSC_COMM_WORLD, error);
 
-    Result result(input, eigenvalues);
-    return result;
+    return std::unique_ptr<AbstractEigenvalueResult<DIM>>(
```
JensWehner, line 360: you can use `make_unique` instead of `new`

applications/DG-Max/DG-Max.cpp
```diff
@@ -188,7 +187,7 @@ int main(int argc, char** argv) {
             points.emplace_back("S");
             points.emplace_back("R");
         }
-        BandstructureGNUPlot<DIM> output(path, points, structure, &result);
+        BandstructureGNUPlot<DIM> output(path, points, structure, &(*result));
```
JensWehner, line 190: why not remove & and * instead? I guess you might have to cast then.

applications/DG-Max/Programs/DGMaxEigen.cpp
```diff
@@ -117,27 +117,23 @@ void runWithDimension() {
...
-        DivDGMaxEigenValue<DIM> solver(*mesh);
         typename DivDGMaxDiscretization<DIM>::Stab stab =
             parsePenaltyParmaters<DIM>();
-        typename DivDGMaxEigenValue<DIM>::Result result =
-            solver.solve(input, stab, p.getValue());
-        if (Base::MPIContainer::Instance().getProcessorID() == 0) {
-            result.printFrequencies();
-            result.writeFrequencies("frequencies.csv");
-        }
+        DivDGMaxEigenvalue<DIM> solver(*mesh, p.getValue(), stab);
```
JensWehner, line 126: can you change p to a more meaningfull name?

applications/Temp/temp.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include "Base/CommandLineOptions.h"
+#include "Geometry/Mappings/RefinementMapsForCube.h"
+
+auto& meshName = Base::register_argument<std::string>(
+    'm', "meshName", "name of the Mesh file", true);
+
+int main(int argc, char* argv[]) {
+    Base::parse_options(argc, argv);
+
+    auto* configData = new Base::ConfigurationData(1, 0);
```
JensWehner, line 12: Why not use unique_ptr?

applications/DG-Max/Algorithms/DivDGMaxEigenvalue.cpp
Laar: Because that has a nicer interface than vector for this usecase. The basic need is the numbers `offset + (0...n-1)`, where the offset changes often but n remains constant. So I used valarray to change the offset.

Looking at it now I don't think I would do it the same way again. But changing it is also not really worth it in my opinion.

applications/DG-Max/DG-Max.cpp
Laar: The reason is that the plotter at the moment expects a raw pointer, as it should work both with and without results, and now that the results are in a smart pointer they need to be unpacked. This is indeed not a perfect situation, but I'm not sure what the best design is to improve it. So if you have any good suggestions, then please let me know.

applications/Temp/temp.cpp
Laar: While a good idea, it would require significant refactoring of the kernel and other applications. So I would say that it would fall outside the scope of this pull request.

Sidenote, in my opinion the need for this configuration data should be removed completely. However, that can probably only be done by changing quite a bit of the architecture.

applications/DG-Max/Algorithms/DivDGMaxEigenvalue.cpp
JensWehner: okay.

applications/DG-Max/DG-Max.cpp
JensWehner: .get() returns the raw pointer

applications/Temp/temp.cpp
JensWehner: I mean you can wrap the `new` call in a unique_ptr and use the `get()` method to pass the raw pointer along, than you do not have to call `delete` in the end

Laar: Hmm, thinking of it. I'm wondering why it was even allocated via new instead of just put on the stack. Anyways, it should not even be in this pull request.

PR #41, opened 2020-06-09
Formatting fixes to the kernel

Laar:
These fixes are generated by running `make format` on the master branch on my computer.

### DIFF
--- a/kernel/Geometry/PointPhysicalBase.cpp
+++ b/kernel/Geometry/PointPhysicalBase.cpp
@@ -40,61 +40,61 @@
 
 namespace Geometry {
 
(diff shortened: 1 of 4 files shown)


JensWehner:
hmmm this should not happen, which clang-format version did you use?

Laar:
I have version 10.0.0
```
/bin/clang-format --version
clang-format version 10.0.0
```

Kind of strange that it now starts complaining about different lines in the same files.

JensWehner:
Yes so clang 9 and clang 10 differ in this formatting. So I would fix it to clang-9 atm?

Laar:
Hmm, it seems that indeed installing and using clang-format 9 fixes this.

PR #42, opened 2020-06-10
Dgmax convergence tests

Laar:
First work at including testing the convergence of DGMax.

This is done in two ways:
 - A SelfTest that computes the vacuum spectrum with both DivDGMax & DGMax and comparse it with previous results.
 - The program DGMaxEigenConvergence that can be used to run more general convergence tests.


Laar:
Somehow it fails on MacOS and I don't have a clue why. It tries to run the test, while it is not compiled. Which is strange to me as it is configured by
```
cmake -H. -Bbuild -DhpGEM_USE_MPI=ON -DENABLE_TESTING=ON -DCMAKE_BUILD_TYPE=Release -DhpGEM_USE_METIS=ON -DhpGEM_USE_COMPLEX_PETSC=ON
```
while the test is guarded by 
```
if (hpGEM_USE_COMPLEX_PETSC AND hpGEM_USE_SLEPC)
```
so I would expect that it would not compile nor run the test case.

JensWehner:
could you reformat it first? It makes it easier to see the proper diff.

applications/DG-Max/Utils/HomogeneousBandStructure.h
```diff
@@ -19,7 +19,15 @@ class HomogeneousBandStructure : public BandStructure<DIM> {
         std::array<LinearAlgebra::SmallVector<DIM>, DIM> reciprocalVectors,
         double permittivity = 1);
 
-    /// \brief Compute the spectrum at a given point in k-space
+    /// \brief Compute the spectrum at a given point in k space
+    ///
+    /// \param kpoint The k-point
+    /// \param omegaMax  The maximum frequency
+    /// \return The frequencies in increasing order at the k-point up to omegaMax.
+    virtual std::vector<double> computeLinearSpectrum(
```
JensWehner, line 27: virtual is not needed here because it is final

JensWehner:
Could the cmake problem be related to this 

> Local Variables CMake has the concept of scope; you can access the value of the variable after you set it as long as you are in the same scope. If you leave a function or a file in a sub directory, the variable will no longer be defined."

Laar:
I found the problem. The `MasterMake.cmake` file apparently configures all the tests based on glob-ing the local files. So as that was included the tests were created, but the corresponding executables were not created as they were excluded by the if-statement.

I don't know whether the MasterMake is still needed if we switch to some testing framework. But if it is then it is probably good to rename it.

applications/DG-Max/Programs/DGMaxEigenConvergence.cpp
```diff
@@ -0,0 +1,205 @@
...
+    if (structureIndex == 0) {
+        std::array<LinearAlgebra::SmallVector<DIM>, DIM> reciprocalVectors;
+        for (std::size_t i = 0; i < DIM; ++i) {
+            for (std::size_t j = 0; j < DIM; ++j) {
+                reciprocalVectors[i][j] = 0;
+            }
+            // Assuming unit cube as unit cell
+            reciprocalVectors[i][i] = 2 * M_PI;
+        }
+        return std::unique_ptr<BandStructure<DIM>>(
```
JensWehner, line 83: preferably make_unique, there is also a clang-tidy workflow for it.

tests/self/DGMax/DivDGmax_Vacuum2D_SelfTest.cpp
```diff
@@ -0,0 +1,94 @@
...
+    //    // Standard cubic lattice.
+    //    reciprocal[0] = LinearAlgebra::SmallVector<2>({2 * M_PI, 0});
+    //    reciprocal[1] = LinearAlgebra::SmallVector<2>({0, 2 * M_PI});
+    //    HomogeneousBandStructure<2> analytical(reciprocal);
+    //    std::vector<double> linear =
+    //        analytical.computeLinearSpectrum(testPoint.getKPoint(), 20);
+    //    result.printFrequencyTable(linear);
+    //    result.printErrorTable(linear);
+    //
+    //    result.printResultCode(10);
```
JensWehner, line 93: do you want to remove the commented out code?

applications/DG-Max/Programs/DGMaxEigenConvergence.cpp
Laar: I have fixed these now, but unfortunately I'm still messing with getting the clang-tidy options into clion. Do you think it would be worth it if we create a .clang-tidy for the project and enforce it?

tests/self/DGMax/DivDGmax_Vacuum2D_SelfTest.cpp
Laar: It can now be activated by a flag, so no.

Laar:
> Really nice, could you add a small Readme.md to explain the idea of the tests a bit more?

I don't know if a readme is not a bit overkill for these tests. They are basically the same convergence test as you would find in any finite element paper. Though in a paper you publish the errors and the convergence rate. Theory then tells you the converge rate should be say 4, and you see if the numbers are close enough. 

Trouble with making tests out of it is that 'close enough' is not really well defined for a computer to check (especially for DGMax, where you have a random amount of extra 0 eigenvalues). So instead of coding 'close enough', we run the tests once, check the correct convergence of the output to theory and put them as expected values. If the code changes the results might change, then you have to manually check the convergence of the new values and put them in the test.

Is this perfect, no, but I don't have a better & stable approach.

Note that the same approach is taken by several tests in the PETSc test directory.

JensWehner:
Your comment makes for a brilliant Readme file.

Laar:
As that is completely separate from this branch I have split it of into #46

.github/workflows/PETSC_complex.yml
```diff
@@ -27,4 +27,4 @@ jobs:
     - name: make
       run: cmake --build build -- -j4
     - name: test
-      run: cd build && ctest
+      run: cd build && ctest -V -R DGmax_Vacuum2D_SelfTest && ctest
```
JensWehner, line 30: sure about this in general?

applications/DG-Max/Utils/BraggStackBandstructure.cpp
```diff
@@ -48,7 +48,13 @@ std::map<double, std::size_t> BraggStackBandstructure::computeSpectrum(
     }
     std::sort(freqs.begin(), freqs.end());
     // Binning
```
JensWehner, line 50: remove comment

Laar: Done

JensWehner (approved):

PR #43, opened 2020-06-15
Small fix to SparsityEstimator

Laar:
This was obviously a typo in the original version.


JensWehner:
We should actually write tests for these cases.

JensWehner:
sth is still wrong with the formatting

PR #44, opened 2020-06-15
fixing clang-format version

JensWehner:


Laar:
On further inspection we might have a bigger problem at hand with the clang-format version. Homebrew only ships version 10 for mac OS. So from a usability point of view it might be better to switch to version 10 if possible.

Laar:
The CI formatter seems to fail with
```
Run sudo apt install libblas-dev liblapack-dev libopenmpi-dev clang-format-10
  sudo apt install libblas-dev liblapack-dev libopenmpi-dev clang-format-10
  shell: /bin/bash -e {0}

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package clang-format-10
##[error]Process completed with exit code 100.
```

Laar:
Thanks for this fix/upgrade. I've tested that it works out of the box on both my (arch) linux & mac.

PR #47, opened 2020-07-09
updated Header guards

JensWehner:
now all header guards have the same format, which should make include conflicts less likely. 

for kernel:

HPGEM_KERNEL_FILENAME_H

for apps:

HPGEM_APP_FILENAME_H


Laar:
Is this not risky with when the same filename would occur in two different context?

JensWehner:
I thought about that, it would matter if the applications can include each other, but that cannot happen atm?. Otherwise it is a lot more secure actually, because now you can have the same filename in kernel and in application, which was not guaranteed before.

JensWehner:
and it should also not interfere with external libraries as well.

Laar:
Was thinking more about internal conflicts, but those should not happen with some good design.

Issue #4, opened 2020-01-07
Introduce a hpGEM namespace


Laar:
We are currently using several rather generic top level namespaces ('Base', 'Geometry', etc.). For use as a library (and to prevent clashes with others) it would be better if we have one top level namespace, say hpgem, and then nest the kernel components in this namespace.

PR #48, opened 2020-07-09
Namespace

JensWehner:
added a hpgem namespace to hpgem,

all applications and tests have a `using namespace hpgem;`. This is not brilliant, but also not worse than before and now at least the lib does not spill out anymore. 

fixes https://github.com/hpgem/hpgem/issues/4


kernel/LinearAlgebra/SmallMatrix_impl.h.orig
```diff
@@ -0,0 +1,352 @@
+/*
```
Laar, line 1: I doubt this should be here

JensWehner: gone

PR #49, opened 2020-07-10
Catch2

JensWehner:
Convert the testsuite to the catch2 framework. 

This does not utliize the whole power of the framework yet, as this is basically just a regex replace. 

At least for the moment I think this is a big enough improvement to merit merging into master.


tests/unit/Utilities/010BasisFunctions_UnitTest.cpp
```diff
@@ -79,18 +85,17 @@ int main() {
 
             point1D[0] += -1e-8;
             ret = test->evalDeriv((point1D));
-            logger.assert_always((std::abs(ret[0] - 5.e7 * (x1 - x0)) < 1e-5),
-                                 "gradient");
-            logger.assert_always((std::abs(test->evalDeriv0((point1D)) -
-                                           5.e7 * (x1 - x0)) < 1e-5),
-                                 "derivative");
+            INFO("gradient");
+CHECK((std::abs(ret[0] - 5.e7 * (x1 - x0)) < 1e-5));
```
Laar, line 89: This seems to be badly indented. Is that a requirement that the CHECK is at the start of the line?

JensWehner: We can also require that we get the local catch2, but all the projects I have seen just put the catch header inside.

JensWehner: @Laar the intendaton is fixed.

JensWehner:
@Laar please review.

tests/unit/Geometry/020PointReference_UnitTest.cpp
```diff
@@ -1,236 +0,0 @@
-/*
```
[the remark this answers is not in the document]
JensWehner: the whole test was commented out, before I got there. So I deleted it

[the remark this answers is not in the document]
Laar: Hmm, I see. Strange.

PR #51, opened 2020-07-14
make format and added a small header to eigenvalue output

JensWehner:


applications/DG-Max/ProblemTypes/AbstractEigenvalueResult.h
```diff
@@ -73,6 +73,7 @@ class AbstractEigenvalueResult {
         for (std::size_t i = 0;
              i < originalProblem().getPath().totalNumberOfSteps(); ++i) {
             std::vector<double> freqs = frequencies(i);
+            stream<<"k-point frequencies->\n";
```
Laar, line 76:
```suggestion
            stream<<"k-point" << separator <<  "frequencies->\n";
```
would keep using the same separator as in the rest of the file

PR #53, opened 2020-08-28
Fix the labeling of subdomain boundary faces

Laar:
The current check is to look if the same (by id) global nodes are both
side of the face. This identical to checking their coordinates but
independent of the dimension. This differs from the previous version
where no only the node coordinates were checked, but also their
ordering (which does not need to be the same on both sides).


kernel/Base/Face.cpp
```diff
@@ -89,71 +89,27 @@ Face::Face(Element* ptrElemL, const std::size_t& localFaceNumberL,
...
+                localFaceNumberL);
+        std::vector<std::size_t> globalRightNodes =
+            ptrElemR->getPhysicalGeometry()->getGlobalFaceNodeIndices(
+                localFaceNumberR);
+        std::sort(globalLeftNodes.begin(), globalLeftNodes.end());
+        std::sort(globalRightNodes.begin(), globalRightNodes.end());
+        if (globalLeftNodes != globalRightNodes) {
+            setFaceType(Geometry::FaceType::PERIODIC_BC);
+        }
+    }
```
JensWehner, line 107: this should be a function.

JensWehner (changes requested):
Nice change, makes the code more readable, but it should still be a function.

JensWehner (changes requested):
Nice, changes, make the code more readable. Can you refactor it into a function

Laar:
Part of the reason for leaving it inline was that the function does not determine what type of periodic boundary face it is, i.e., subdomain or not. Currently I'm not wanting to change too much to prevent it accidentally breaking.

I hope that if we refactor the Mesh that this will be improved a lot.

kernel/Base/Face.cpp
```diff
@@ -364,71 +308,15 @@ void Face::addElement(Element* ptrElementR, std::size_t localFaceNumberR) {
     std::vector<std::size_t> localRightNodes =
         ptrElementR->getPhysicalGeometry()->getLocalFaceNodeIndices(
             localFaceNumberR);
+
+    if (detirmineIsPeriodicBoundaryFace()) {
```
JensWehner, line 312: fix spelling pls.
```suggestion
    if (determineIsPeriodicBoundaryFace()) {
```

kernel/Base/Face.cpp
```diff
@@ -89,71 +89,15 @@ Face::Face(Element* ptrElemL, const std::size_t& localFaceNumberL,
     std::vector<std::size_t> localRightNodes =
         ptrElemR->getPhysicalGeometry()->getLocalFaceNodeIndices(
             localFaceNumberR);
+
+    if (detirmineIsPeriodicBoundaryFace()) {
```
JensWehner, line 93:
```suggestion
    if (determineIsPeriodicBoundaryFace()) {
```

kernel/Base/Face.cpp
```diff
@@ -455,6 +343,21 @@ const std::vector<Base::Node*> Face::getNodesList() const {
 
     return ptrNodesAtFace;
 }
+
+bool Face::detirmineIsPeriodicBoundaryFace() {
```
JensWehner, line 347:
```suggestion
bool Face::determineIsPeriodicBoundaryFace() {
```

kernel/Base/Face.h
```diff
@@ -377,6 +377,10 @@ class Face final : public Geometry::FaceGeometry, public FaceData {
 
     std::vector<std::size_t> numberOfConformingDOFOnTheFace_;
     std::size_t faceID_;
+
+    /// Determine whether this face is a face on the periodic boundary. Is
+    /// independent of the faceType.
+    bool detirmineIsPeriodicBoundaryFace();
```
JensWehner, line 383: should be const 
```suggestion
    bool determineIsPeriodicBoundaryFace()const;
```

JensWehner (changes requested):
@Laar good refactor, just a small spelling mistake and `const` stuff

Laar:
> @Laar good refactor, just a small spelling mistake and `const` stuff

@JensWehner Done, needed a little more formatting (and the `const` in the cpp file).

JensWehner (approved):
LGTM

PR #54, opened 2020-08-28
Overhaul DGMax Eigenvalue solver

Laar:
Includes: #42 and #53 

A major overhaul of the Eigenvalue solver using the DGMax discretization. The main components of this overhaul are:
 - The introduction of several helper classes for working with the k-phase shifted matrices
 - The introduction of a Projector for the kernel of the stiffness matrix
 - Restructuring the implementation of DGmaxEigenvalue for the added complexity

The required new helper classes are `MatrixBlocks` and `KPhaseShift`. These help implementing the stiffness matrix and the matrix for the projector, which are both parameterized based on the wavevector k. This parameterization takes the form that some entries (those corresponding to the boundary of the mesh) are of the form `A(k)_{jl} = exp(i (k . x_jl)` where x_jl is the distance between the two periodic boundary faces. MatrixBlocks represents a block in a spare matrix, KPhaseShift uses such blocks in combination with the value of x_{jl} to phase shift the entries.

The projector is based on that the kernel of the stiffness matrix are the gradients of conforming Lagrange elements. This results in a matrix B with a non orthogonal basis of the kernel, i.e., `SB = 0`), in code the matrix B* is implemented. Using this we can construct a projector `P = I - B* (BB*)^{-1}B` which can then be used in the implementation. Unfortunately, it seems that the gain from this projector is minor.

PR #56, opened 2020-10-21
Dgmax null space polishing

Laar:
Polishing based on #54 that is not in the scope of that branch.

### DIFF
--- a/applications/DG-Max/DG-Max.cpp
+++ b/applications/DG-Max/DG-Max.cpp
@@ -124,20 +124,31 @@ int main(int argc, char** argv) {
         ///////////////////
         // Harmonic code //
         ///////////////////
+        for (auto* element : mesh->getElementsList()) {
+            element->setNumberOfTimeIntegrationVectors(3);
+        }
 
-        //        DGMaxHarmonic harmonicSolver (base, p.getValue());
-        //        DivDGMaxHarmonic harmonicSolver (base);
-        //
-        //        SampleHarmonicProblems problem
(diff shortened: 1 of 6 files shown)


applications/DG-Max/DG-Max.cpp
```diff
@@ -124,20 +124,31 @@ int main(int argc, char** argv) {
...
+        //                  << std::endl;
+        //        std::cout << "HCurl error " <<
+        //        errors[DGMaxDiscretizationBase::HCurl]
+        //                  << std::endl;
+        // Seems to be broken :(
+        //        DivDGMaxHarmonic<DIM> harmonicSolver(*mesh);
+        //        harmonicSolver.solve(harmonicProblem, divStab, p.getValue());
+        //        std::cout << "L2 error "
+        //                  << harmonicSolver.computeL2Error(harmonicProblem)
+        //                  << std::endl;
```
JensWehner, line 149: pls remove

JensWehner (changes requested):
Just remove commented out code. Otherwise nice to put the input functions into a map.

applications/DG-Max/DG-Max.cpp
Laar: I do like to keep it. This program is like the debugging/testing program. Thus I would like to be able to have a sample piece of code for each of the implementations that I can easily switch to.

A better solution may be possible, but is in my opinion not worth the time at the moment.

JensWehner:
format is failing atm.

PR #58, opened 2020-11-24
Let the preprocessor show some basic mesh statistics

Laar:
This is useful to view at a generated mesh to get a basic overview of its properties.


JensWehner (approved):
Nice work.

PR #59, opened 2020-11-24
Higher order 2dnedelec

Laar:
This replaces the implementation Nédélec basis functions on a triangle so that we can support more than just the lowest order.

These basis functions are used in DGMax, and we currently only support higher order Nédélec functions on tetrahedra. Adding higher order ones in 2D means that we can test codes with higher order basis functions in 2D instead of 3D. This is much cheaper from a computational point of view.

### DIFF
--- a/kernel/Utilities/BasisFunctions1DH1ConformingLine.cpp
+++ b/kernel/Utilities/BasisFunctions1DH1ConformingLine.cpp
@@ -72,17 +72,27 @@ double BasisFunction1DInteriorLine::evalDeriv0(
                LobattoPolynomialDerivative(polynomialOrder_, p[0]) / 4.;
 }
 
-Base::BasisFunctionSet* createDGBasisFunctionSet1DH1Line(
+std::vector<Base::BaseBasisFunction*> createDGBasisFunctions1DH1Line(
     std::size_t polynomialOrder) {
-    Base::BasisFunctionSet* result(new Base::BasisFunctionSet(polynomialOrder));
+    std::vector<Base::BaseBasisFunction*> result;
     if (polynomialOrder > 0) {
-        result->addBasisFunction(new BasisFunction1DVertexLine(0));
-        result->addBasisFunction(new BasisFunction1DVertexLine(1));
+        result.emplace_back(new BasisFunction1DVertexLine(0));
+        result.emplace_back(new BasisFunction1DVertexLine(1));
         for (std::size_t i = 0; i + 2 <= polynomialOrder; ++i) {
-            result->addBasisFunction(new BasisFunction1DInteriorLine(i));
+            result.emplace_back(new BasisFunction1DInteriorLine(i));
         }
     } else {
-        addPiecewiseConstantBasisFunction1D(*result);
+        result.emplace_back(createPiecewiseConstant1D());
+    }
+    return result;
+}
+
+Base::BasisFunctionSet* createDGBasisFunctionSet1DH1Line(
+    std::size_t polynomialOrder) {
+
+    Base::BasisFunctionSet* result(new Base::BasisFunctionSet(polynomialOrder));
+    for (auto* bf : createDGBasisFunctions1DH1Line(polynomialOrder)) {
+        result->addBasisFunction(bf);
     }
     return result;
 }
--- a/kernel/Utilities/BasisFunctions2DH1ConformingTriangle.cpp
+++ b/kernel/Utilities/BasisFunctions2DH1ConformingTriangle.cpp
@@ -173,29 +173,38 @@ double BasisFunction2DInteriorTriangle::evalDeriv1(
                     LobattoPolynomialDerivative(polynomialOrder1_, x1));
 }
 
-Base::BasisFunctionSet* createDGBasisFunctionSet2DH1Triangle(
+std::vector<Base::BaseBasisFunction*> createDGBasisFunctions2DH1Triangle(
     std::size_t order) {
-    Base::BasisFunctionSet* result(new Base::BasisFunctionSet(order));
+    std::vector<Base::BaseBasisFunction*> result;
     if (order > 0) {
         for (std::size_t i = 0; i < 3; ++i) {
-            result->addBasisFunction(new BasisFunction2DVertexTriangle(i));
+            result.emplace_back(new BasisFunction2DVertexTriangle(i));
         }
         for (std::size_t k = 0; k + 2 <= order; ++k) {
             for (std::size_t i = 0; i < 3; ++i) {
                 for (std::size_t j = 0; j < i; ++j) {
-                    result->addBasisFunction(
+                    result.emplace_back(
                         new BasisFunction2DFaceTriangle(i, j, k));
                 }
             }
             if (k + 3 <= order) {
                 for (std::size_t j = 0; j <= k; ++j) {
-                    result->addBasisFunction(
+                    result.emplace_back(
                         new BasisFunction2DInteriorTriangle(k - j, j));
                 }
             }
         }
     } else {
-        addPiecewiseConstantBasisFunction2D(*result);
+        result.emplace_back(createPiecewiseConstant2D());
+    }
+    return result;
+}
+
+Base::BasisFunctionSet* createDGBasisFunctionSet2DH1Triangle(
+    std::size_t order) {
+    Base::BasisFunctionSet* result(new Base::BasisFunctionSet(order));
+    for (auto* bf : createDGBasisFunctions2DH1Triangle(order)) {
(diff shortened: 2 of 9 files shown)


kernel/Utilities/BasisFunctions1DH1ConformingLine.cpp
```diff
@@ -72,17 +72,27 @@ double BasisFunction1DInteriorLine::evalDeriv0(
...
+
+Base::BasisFunctionSet* createDGBasisFunctionSet1DH1Line(
+    std::size_t polynomialOrder) {
+
+    Base::BasisFunctionSet* result(new Base::BasisFunctionSet(polynomialOrder));
+    for (auto* bf : createDGBasisFunctions1DH1Line(polynomialOrder)) {
+        result->addBasisFunction(bf);
     }
     return result;
 }
```
JensWehner, line 98: we could start introducing unique_ptrs

kernel/Utilities/BasisFunctions2DH1ConformingTriangle.cpp
```diff
@@ -173,29 +173,38 @@ double BasisFunction2DInteriorTriangle::evalDeriv1(
                     LobattoPolynomialDerivative(polynomialOrder1_, x1));
 }
 
-Base::BasisFunctionSet* createDGBasisFunctionSet2DH1Triangle(
+std::vector<Base::BaseBasisFunction*> createDGBasisFunctions2DH1Triangle(
     std::size_t order) {
-    Base::BasisFunctionSet* result(new Base::BasisFunctionSet(order));
+    std::vector<Base::BaseBasisFunction*> result;
     if (order > 0) {
         for (std::size_t i = 0; i < 3; ++i) {
```
JensWehner, line 180: what does the `3` relate to?

kernel/Utilities/BasisFunctions2DH1ConformingTriangle.cpp
```diff
@@ -173,29 +173,38 @@ double BasisFunction2DInteriorTriangle::evalDeriv1(
...
             }
         }
     } else {
-        addPiecewiseConstantBasisFunction2D(*result);
+        result.emplace_back(createPiecewiseConstant2D());
+    }
+    return result;
+}
+
+Base::BasisFunctionSet* createDGBasisFunctionSet2DH1Triangle(
```
JensWehner, line 203: I do not fully get why we return this as a pointer, can it be a derived class?

kernel/Utilities/BasisFunctions2DNedelec.cpp
```diff
@@ -37,77 +37,229 @@
...
+///
+/// The function F(x,y) gives the direction of the vector field and consists of
+/// two linear functions. For each side these are constructed so that it has a
+/// constant tangential trace along that side and zero tangential trace along
+/// the two other sides.
+///
+/// The function q_p(t) is shared between the sides and 'modulates' the vector
+/// field by a polynomial of order p-1. The coordinate t is linear on the side
+/// to which the basis function is associated.
+class BasisCurlEdgeNedelec2D : public Base::BaseBasisFunction {
```
JensWehner, line 65:
```suggestion
class BasisCurlEdgeNedelec2D final: public Base::BaseBasisFunction {
```

or can it be derived from

kernel/Utilities/BasisFunctions2DNedelec.cpp
```diff
@@ -37,77 +37,229 @@
...
+   public:
+    BasisCurlEdgeNedelec2D(
+        std::size_t side,
+        std::shared_ptr<const Base::BaseBasisFunction> modulator)
+        : side_(side), modulator_(std::move(modulator)) {
```
JensWehner, line 70: why do you move here? If you want to move take modulator as &&

kernel/Utilities/BasisFunctions2DNedelec.cpp
```diff
@@ -37,77 +37,229 @@
...
+        coord[0] = evalModulatorCoord(p);
+        ret *= modulator_->eval(coord);
+    }
+
+    LinearAlgebra::SmallVector<2> evalCurl(
+        const Geometry::PointReference<2>& p) const override;
+
+   private:
+    /// The side function F(x,y)
+    void evalSideFunction(const Geometry::PointReference<2>& p,
```
JensWehner, line 88: probably because od the interface but it would be much nicer to return ret instead of passing it in as a non-const ref

kernel/Utilities/BasisFunctions2DNedelec.cpp
```diff
@@ -37,77 +37,229 @@
...
+    double result;
+
+    Geometry::PointReference<1> modulatorCoord;
+    modulatorCoord[0] = evalModulatorCoord(p);
+    double modulatorVal = modulator_->eval(modulatorCoord);
+    LinearAlgebra::SmallVector<2> sideVal;
+    evalSideFunction(p, sideVal);
+
+    // Compute rotation Fy/dx - Fx/dy from the side function
+    result = -2.0;
```
JensWehner, line 133:
```suggestion
    double result = -2.0;
```

kernel/Utilities/BasisFunctions2DNedelec.cpp
```diff
@@ -37,77 +37,229 @@
...
+    // Side of the triangle to which this basis function is associated
+    std::size_t side_;
+    // 1D Modulation function for the basis functions.
+    std::shared_ptr<const Base::BaseBasisFunction> modulator_;
+};
+
+LinearAlgebra::SmallVector<2> BasisCurlEdgeNedelec2D::evalCurl(
+    const Geometry::PointReference<2>& p) const {
+
+    double result;
```
JensWehner, line 124:
```suggestion

```

JensWehner (commented):
I do not like all the pointers otherwise it seems good.

kernel/Utilities/BasisFunctions1DH1ConformingLine.cpp
Laar: That might result in quite a bit of refactoring. As that would impact all the basis functions, as these are passed around to the mesh. Preferably I would redesign this basis function stuff at some point.

All the changes in this file and the one for H1ConformingTriangle, are just so that I can reuse these basis functions as building block (the modulator) for the 2D Nedelec ones. So I preferably do as little as possible in these files.

kernel/Utilities/BasisFunctions2DH1ConformingTriangle.cpp
Laar: Here I'm not completely sure either. But there are places where we use a derived class (has to do with face orientation). So I guess to keep a consistent interface between all the different basis function types.

kernel/Utilities/BasisFunctions2DNedelec.cpp
Laar: I need multiple copies of the shared pointer. So I probably should have passed it by reference and let it make a local copy. Will change that.

kernel/Utilities/BasisFunctions2DNedelec.cpp
Laar: I would keep it consistent with all the interface for `eval`, where it is also passed as out-parameter.

kernel/Utilities/BasisFunctions1DH1ConformingLine.cpp
JensWehner: yes okay

kernel/Utilities/BasisFunctions2DH1ConformingTriangle.cpp
JensWehner: okay

kernel/Utilities/BasisFunctions2DH1ConformingTriangle.cpp
JensWehner: can you turn it into a variable?

kernel/Utilities/BasisFunctions2DNedelec.cpp
JensWehner: okay

kernel/Utilities/BasisFunctions2DH1ConformingTriangle.cpp
Laar: Could, but to do that nicely I need to go through `Geometry::ReferenceTetrahedron::Instance().numberOfVertices()` or so. Which to me feels not like making it much nicer.

JensWehner (approved):

PR #62, opened 2020-12-02
Kshift divdgmax eigenvalue

Laar:
Replace the manual phase shifts in `DivDGMaxEigenvalue` with the abstracted approach from `KPhaseShift`.

As can be seen in the diff, this removes a lot of complicated index-fiddling in DivDGMaxEigenvalue.


JensWehner (approved):
Ohh that is a nice refactor. Well done

PR #63, opened 2020-12-03
Localindexing

Laar:
Introduce `LocalIndexing`, an abstraction around the ordering of the basisfunctions with support on a element or face.

So far it is not used, but it could, for example, replace `FaceDoFInfo` in `DivDGMaxDiscretization.cpp`. It is being used in some side projects (that are not ready at all).


tests/unit/Utilities/001LocalIndexing_UnitTest.cpp
```diff
@@ -0,0 +1,144 @@
...
+    REQUIRE(indexing.getNumberOfDoFs() == 0);
+}
+
+int main(int argc, char* argv[]) {
+    // Only pass program name
+    Base::parse_options(1, argv);
+
+    int result = Catch::Session().run(argc, argv);
+
+    return result;
```
JensWehner, line 143: What is the intent behind this?

Laar: Making sure that MPI is initialized if needed. Without it there are some internals that will crash due to initialized MPI (and possibly other parts). However, I do want to pass the actual program arguments to catch.

Laar: I've replaced to comment with a more clarifying one.

kernel/Utilities/ElementLocalIndexing.cpp
```diff
@@ -0,0 +1,129 @@
...
+        for (std::size_t unknown : includedUnknowns_) {
+            std::size_t numDoFs = element->getNumberOfBasisFunctions(unknown);
+            offsets_[unknown] = totalNumberOfDofs_;
+            sizes_[unknown] = numDoFs;
+            totalNumberOfDofs_ += numDoFs;
+        }
+    }
+}
+
+void ElementLocalIndexing::validate() const {
```
JensWehner, line 100: I am not sure what the idea behind this function is can you turn `validate` into a more descriptive name?

kernel/Utilities/ElementLocalIndexing.h
```diff
@@ -0,0 +1,152 @@
...
+    /// Get the current element.
+    const Base::Element* getElement() const { return element_; }
+    /// The unknowns that are included.
+    const std::vector<std::size_t>& getIncludedUnknowns() const {
+        return includedUnknowns_;
+    }
+
+    /// Check the invariants of the datastructure to see if the internal
+    /// constraints are satisfied. To be used for testing and debugging.
+    void validate() const;
```
JensWehner, line 116: I am not a big fan of exposting API functions that are only used for testing.

kernel/Utilities/ElementLocalIndexing.h
```diff
@@ -0,0 +1,152 @@
...
+        result.reinit(unknowns);
+        return result;
+    }
+
+   private:
+    /// The current element
+    const Base::Element* element_;
+
+    /// The maximum number of unknowns.
+    std::size_t numberOfUnknowns_;
```
JensWehner, line 135: is that not simply `sizes_.size()`

kernel/Utilities/ElementLocalIndexing.h
```diff
@@ -0,0 +1,152 @@
...
+
+    /// The total number of DoFs of the included unknowns.
+    std::size_t totalNumberOfDofs_;
```
JensWehner, line 138: is that not simply the sum of `sizes_`

kernel/Utilities/FaceLocalIndexing.h
```diff
@@ -0,0 +1,125 @@
...
+}
+
+namespace Utilities {
+
+/**
+ * Index of the local DoFs on a Face.
+ *
+ * For face matrices and vectors the rows and columns correspond to DoFs of a
+ * particular (sub)set of unknowns. These are the DoFs which have support on the
+ * adjacent element(s). This calss provides easy access to information about the
```
JensWehner, line 56:
```suggestion
 * adjacent element(s). This class provides easy access to information about the
```

kernel/Utilities/FaceLocalIndexing.h
```diff
@@ -0,0 +1,125 @@
...
+        }
+    }
+
+    /// \return The total number of DoFs for all the included unknowns.
+    std::size_t getNumberOfDoFs() const {
+        return left_.getNumberOfDoFs() + right_.getNumberOfDoFs();
+    }
+
+    /// \return The unknowns that are included.
+    const std::vector<std::size_t>& getIncludedUnknowns() {
```
JensWehner, line 107: should be a const function?

kernel/Utilities/FaceLocalIndexing.h
```diff
@@ -0,0 +1,125 @@
...
+        return left_.getIncludedUnknowns();
+    }
+
+    /// \return The face for which this indexing is currently.
+    const hpgem::Base::Face* getFace() { return face_; }
```
JensWehner, line 112: same here

tests/unit/Utilities/001LocalIndexing_UnitTest.cpp
```diff
@@ -0,0 +1,147 @@
...
+int main(int argc, char* argv[]) {
+
+    // The mesh reader needs an initialized MPI and possibly more. To achieve
+    // this initialize the system using only the program name, as the arguments
+    // to the program are most probably used to direct Catch.
+    Base::parse_options(1, argv);
+
+    int result = Catch::Session().run(argc, argv);
+
+    return result;
```
JensWehner, line 146: I do not fully understand why this is necessary.

JensWehner (changes requested):
I do not fully understand the idea behind this class, but it seems to be a lack of domain knowledge on my part. I think a couple of more `const` could help.

kernel/Utilities/ElementLocalIndexing.h
Laar: Indeed, replaced it.

kernel/Utilities/ElementLocalIndexing.h
Laar: Yes, and also to `offsets_.size()`, but both are far less descriptive in my opinion.

kernel/Utilities/FaceLocalIndexing.h
Laar: Done

kernel/Utilities/FaceLocalIndexing.h
Laar: Done

kernel/Utilities/ElementLocalIndexing.cpp
Laar: Renamed

kernel/Utilities/ElementLocalIndexing.h
Laar: For me it depends. I have moved most functionality out of this method to the actual test. But there are several cases where (especially in concurent/MPI settings), where in debugging cases it is easy to test whether the state is valid. Moreover, this way ensures that the internal invariants are stated somewhere.

kernel/Utilities/ElementLocalIndexing.h
Laar: It's now gone entirely and the size of `offsets_` and `sizes_` is now dynamical (i.e. sufficient to for the largest included unknown).

tests/unit/Utilities/001LocalIndexing_UnitTest.cpp
Laar: There are some things inside the mesh reader (that I use for creating valid elements), that depend on the processor rank (index) in an MPI context. MPI requires initialization before any function, including the functions that ask for the processor rank.

The mesh probably needs only MPI, but I'm not certain that that is the case. So this is the way to initialize the whole DGMax properly.

tests/unit/Utilities/001LocalIndexing_UnitTest.cpp
JensWehner: okay nice

kernel/Utilities/ElementLocalIndexing.h
JensWehner: okay

tests/unit/Utilities/001LocalIndexing_UnitTest.cpp
JensWehner: I find it a bit weird, why not intialized MPI inside the test?

kernel/Utilities/ElementLocalIndexing.cpp
```diff
@@ -95,33 +95,14 @@ void ElementLocalIndexing::reinit(const Base::Element *element) {
     }
 }
 
-void ElementLocalIndexing::validate() const {
+void ElementLocalIndexing::validateInternalState() const {
```
JensWehner, line 97: this is a much nicer name. Thanks

kernel/Utilities/ElementLocalIndexing.cpp
```diff
@@ -0,0 +1,107 @@
...
+    }
+    // Check maximum and extend if necessary
+    if (!includedUnknowns_.empty()) {
+        std::size_t maxUnknown =
+            includedUnknowns_[includedUnknowns_.size() - 1];
+        if (maxUnknown >= sizes_.size()) {
+            sizes_.resize(maxUnknown + 1, 0);
+            offsets_.resize(maxUnknown + 1, -1);
+        }
+    }
```
JensWehner, line 69: Should this be outside of the ifdef?

kernel/Utilities/ElementLocalIndexing.cpp
```diff
@@ -0,0 +1,107 @@
...
+#endif
+    // Update the content
+    if (element_ != nullptr) {
+        // Clear the storage to the default values as reinit does only overwrite
+        // the included unknowns.
+        std::fill(offsets_.begin(), offsets_.end(), UNKNOWN_NOT_INCLUDED);
+        std::fill(sizes_.begin(), sizes_.end(), 0);
```
JensWehner, line 76: could make that a function

JensWehner (approved):
Looking good, there are some small things you could change. Merge at your own discretion. :+1:

tests/unit/Utilities/001LocalIndexing_UnitTest.cpp
Laar: Because you only need to do it once. 

So if we later add multiple tests, then which runs first? Or if you specify that it only needs to run one of the two? From that point of view I found it more logical to set it up before anything else.

kernel/Utilities/ElementLocalIndexing.cpp
Laar: Oops, good catch!

PR #64, opened 2020-12-07
Dgmax eigenvalue driver

Laar:
Replace the way the solvers for the eigenvalue problems in DGMax are controlled.

The previous approach used a fixed path (`KSpacePath`) and after solving at each k-point the problem, returned the frequencies at each point. This approach is very inflexible for future extension. All options need to be set beforehand and results can only be viewed after solving the eigenvalue problem at all wave vectors (k points) in the path.

This PR introduces a new, more 'interactive', approach. The eigenvalue solver is controlled by a driver. This driver tells the eigenvalue at what wave vector to solve the problem. The eigenvalue solver then solves it and reports back with the results. The driver can then decide what to do with these results and what is the next wave vector to solve at.

The current result class is very small, exposing only the resulting eigenfrequencies. This is intentional to keep the PR manageable. Exposing more information about the result (e.g. the field) will be part of future PRs.

Reviewing notes:
 - Start with the trio `AbstractEigenvalueSolver`, `AbstractEigenvalueResult` and `AbstractEigenvalueSolverDriver`. This gives the new interface. The rest of the changes is primarily adapting the rest of the code to this new interface.
 - The commented out code in DivDGMaxEigenvalue was noted, but kept. Removing should be part of a separate PR, as it removes functionality that works but is not beneficial.


applications/DG-Max/Programs/DGMaxEigen.cpp
```diff
@@ -95,6 +96,78 @@ int main(int argc, char** argv) {
...
+
+   public:
+    DGMaxEigenDriver(KSpacePath<DIM>& path,
+                     std::size_t targetNumberOfEigenvalues)
+        : targetNumberOfEigenvalues_(targetNumberOfEigenvalues),
+          currentPoint_(0),
+          path_(path),
+          frequencyResults_(path.totalNumberOfSteps()) {
+
+        outFile.open("frequencies.csv");
```
JensWehner, line 110: do you want a hardcoded filename here?

JensWehner (approved):
Looking good. Nice work.

applications/DG-Max/Programs/DGMaxEigen.cpp
Laar: It was previously also hardcoded, and for me it's nice to have a fixed value (no searching for the relevant file).

On the cluster I always run the programs in a new (empty) working directory, so no risk to have clashes. So at the moment I would say that it would be more work to make it variable, for something that won't be used anywhere soon.

applications/DG-Max/Algorithms/DGMaxEigenvalue.cpp
```diff
@@ -244,13 +250,30 @@ void sortEigenvalues(std::vector<PetscScalar>& result) {
               });
 }
 
+template <std::size_t DIM>
+struct DGMaxEigenvalueResult : public AbstractEigenvalueResult<DIM> {
+    std::vector<double> getFrequencies() override final {
```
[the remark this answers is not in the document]
Laar: Fixed, my Java roots shine through again.

PR #66, opened 2020-12-10
Partial face coupling/Sparsity estimator rewrite

Laar:
Original intention was to add partial face coupling to the `SparsityEstimator`.  This can be used when the rows and columns of a matrix correspond to more than one unknown/variable. In such a configuration it may be that some unknowns do not have a contribution in the face matrix, or only couple to some of the unknowns. The idea is to pass this information to both `GlobalPetscMatrix` and `SparsityEstimator` so that they can use this in their assembly.

Primary motivation for this is that without it we can't construct a `GlobalPetscMatrix` with both conforming and non conforming finite elements while using face matrices (which I need/want). By excluding the conforming unknowns from the face matrix we circumvent the problem, which is still documented in `SparsityEstimator.cpp`. Secondary motivation is that a better estimate will reduce the amount of space needed for the `GlobalPetscMatrix`.

While writing and documenting the first version of this PR, I realized how overly complex `SparsityEstimator` had become. Hence, I rewrote it using a much simpler approach. The approach is documented in detail in the cpp file.

Reviewing advice:
 - `Table2D` with its associated test can be reviewed independently. It basically is just a vector with 2D indexing.
 - `SparsityEstimator.cpp` is probably best reviewed by not looking at the diff and just looking at the new implementation, as almost only the constructor has remained of the previous implementation.
 - The added test for the SparsityEstimator exploits the simple structure of a 1D mesh and the knowledge about the number of basis functions and to which mesh part they associate.
 - Changes to the other self tests were to make the solvers be more stable. Somehow they did not converge anymore, while the matrices were identical.


kernel/Utilities/SparsityEstimator.cpp
```diff
@@ -15,310 +18,416 @@ SparsityEstimator::SparsityEstimator(const GlobalIndexing& rowindexing,
...
-        if (includeFaceCoupling) {
-            for (const Base::Face* face : element->getFacesList()) {
-                if (!face->isInternal()) continue;
-                const Base::Element* other = face->getPtrOtherElement(element);
-                addElementDoFs(other, workspace);
-            }
-        }
+         rowIndexing_.getMesh()->getElementsList(Base::IteratorType::GLOBAL)) {
+        // Face, Edge and Node based DoFs will be used in the element matrices
+        // on the ghost layer. So we need to iterate over all elements and not
```
JensWehner, line 247: what is the ghost layer?

kernel/Utilities/SparsityEstimator.cpp
```diff
@@ -15,310 +18,416 @@ SparsityEstimator::SparsityEstimator(const GlobalIndexing& rowindexing,
...
-                // face of the elements adjacent to the face
-                addElementDoFs(otherElement, workspace);
+                for (Detail::DoFInterval rowInterval : rowIntervals) {
+                    for (Detail::DoFInterval colInterval : ownedColIntervals) {
+                        owned[rowInterval].emplace(colInterval);
+                    }
+                    for (Detail::DoFInterval colInterval :
+                         nonOwnedColIntervals) {
+                        nonOwned[rowInterval].emplace(colInterval);
+                    }
```
JensWehner, line 271: We append all elements in the for loop, can we not simply append the vectors.

`a.insert(std::end(a), std::begin(b), std::end(b));`

kernel/Utilities/SparsityEstimator.cpp
```diff
@@ -15,310 +18,416 @@ SparsityEstimator::SparsityEstimator(const GlobalIndexing& rowindexing,
...
-                addElementDoFs(otherElement, workspace);
-            }
+
+    // If there is no face coupling anywhere (e.g. dg mass matrix) then we can
+    // skip looping over the adjacent elements.
+    bool anyFaceCoupling = false;
+    for (std::size_t i = 0; i < faceCoupling.getSize(); ++i) {
+        if (faceCoupling[i]) {
+            anyFaceCoupling = true;
+            break;
```
JensWehner, line 283: Can we iterate over facecoupling itself instead of using an index?

kernel/Utilities/SparsityEstimator.cpp
```diff
@@ -15,310 +18,416 @@ SparsityEstimator::SparsityEstimator(const GlobalIndexing& rowindexing,
...
+    bool distributedMesh = false;
+
+    if (anyFaceCoupling) {
+        for (const Base::Face* face : rowIndexing_.getMesh()->getFacesList()) {
+            // Check for a distributed mesh
+            if (face->getFaceType() == Geometry::FaceType::SUBDOMAIN_BOUNDARY ||
+                face->getFaceType() ==
+                    Geometry::FaceType::PERIODIC_SUBDOMAIN_BC) {
+                distributedMesh = true;
             }
```
JensWehner, line 296: can we turn this into a small separate function, like `return bool CheckdistributedMesh()` ?

kernel/Utilities/SparsityEstimator.cpp
```diff
@@ -15,310 +18,416 @@ SparsityEstimator::SparsityEstimator(const GlobalIndexing& rowindexing,
...
+                const Detail::DoFIntervals& rowIntervals =
+                    rowDoFs.getOwnedIntervals(rowUnknown);
+
+                for (std::size_t ci = 0;
+                     ci < columnIndexing_.getNumberOfIncludedUnknowns(); ++ci) {
+                    std::size_t colUnknown =
+                        columnIndexing_.getIncludedUnknowns()[ci];
+                    if (!faceCoupling(ri, ci)) {
+                        continue;
+                    }
```
JensWehner, line 315:
```suggestion
                    if (!faceCoupling(ri, ci)) {
                        continue;
                    }
                      std::size_t colUnknown =
                        columnIndexing_.getIncludedUnknowns()[ci];
```

kernel/Utilities/SparsityEstimator.cpp
```diff
@@ -15,310 +18,416 @@ SparsityEstimator::SparsityEstimator(const GlobalIndexing& rowindexing,
...
+
+                    for (Detail::DoFInterval rowInterval : rowIntervals) {
+                        for (Detail::DoFInterval colInterval :
+                             ownedColIntervals) {
+                            owned[rowInterval].emplace(colInterval);
+                        }
+                        for (Detail::DoFInterval colInterval :
+                             nonOwnedColIntervals) {
+                            nonOwned[rowInterval].emplace(colInterval);
+                        }
```
JensWehner, line 330: an insert call should be enough here I think

kernel/Utilities/SparsityEstimator.h
```diff
@@ -42,26 +43,11 @@ class SparsityEstimator {
...
-    /// write the DoF count. \param workspace The workspace with DoF counts
-    /// \param nonZeroPerRowOwned see computeSparsityEstimate(std::vector<int>&,
-    /// std::vector<int>&) \param nonZeroPerRowNonOwned see
-    /// computeSparsityEstimate(std::vector<int>&, std::vector<int>&)
-    template <typename GEOM>
-    void writeDoFCount(const GEOM* geom, const Workspace& workspace,
-                       std::vector<int>& nonZeroPerRowOwned,
-                       std::vector<int>& nonZeroPerRowNonOwned) const;
+    void computeSparsityEstimate(std::vector<int>& nonZeroPerRowOwned,
+                                 std::vector<int>& nonZeroPerRowNonOwned,
```
JensWehner, line 47: I think it is nicer if it returned a std::pair or a struct instead of having two non-const reference arguments.

kernel/Utilities/Table2D.h
```diff
@@ -0,0 +1,149 @@
...
+
+    /// Total number of entries
+    std::size_t getSize() const { return rows_ * columns_; }
+
+    /// Get the value of a certain entry
+    ///
+    /// \param n The row number
+    /// \param m The column number
+    /// \return The value
+    typename std::vector<type>::reference operator()(std::size_t n,
```
JensWehner, line 93: very nice

tests/self/PETSc/007SparsityEstimator_SelfTest.cpp
```diff
@@ -383,6 +400,84 @@ void testConformingWith1DMesh() {
     }
 }
 
+/**
+ * Test with conforming basis functions on a 1D mesh. Here the sparisty pattern
```
JensWehner, line 404:
```suggestion
 * Test with conforming basis functions on a 1D mesh. Here the sparsity pattern
```

JensWehner (commented):
The sparsity estimator is simpler now. I am still struggling with the general structurre of the code. This is still a nice simplification. On some level it would be nice to make Table2D a specialisation of MiddleSizeMatrix, but then we would have to template that.

kernel/Utilities/GlobalMatrix.h
```diff
@@ -118,22 +121,47 @@ class GlobalPetscMatrix : public GlobalMatrix {
...
 
     ~GlobalPetscMatrix() override;
 
     void assemble() override;
 
-    void reinit() override;
+    void reinit() override { reinit(true); }
+    /// Reinitialize the matrix
+    /// \param reinitFaceCoupling Whether to reinit the face coupling
+    void reinit(bool reinitFaceCoupling);
```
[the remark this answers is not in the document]
Laar: I don't think there is any use, you either reinit it to the default value, or you keep the old one. I have no foresight about any future values and the direction, so that would make naming it be quite hard.

So any suggestions would be welcome.

kernel/Utilities/SparsityEstimator.cpp
Laar: In an MPI setting, where you have multiple processes, you distribute the mesh so that each process owns part of the mesh. This ownership then results in a distribution of the work, for example, to assemble a global matrix as combination of element  and face matrices, each process only adds the element matrices for the elements and faces it owns.

Ideally, each process would then load only the part of the mesh that you own. Thereby saving the memory, time, etc. of not loading the mesh parts owned by the other tens of processors. However, only knowing about the mesh the process owns is insufficient to do all its calculations. For example, when constructing a face matrix on the subdomain boundary (=boundary between the ownership region of two processes), you would need to know about the element that is owned by the other process. 

So each processor also loads a small layer of elements that surround the elements that it owns itself. This layer of elements, in our case 1 element wide, is known as the ghost layer. The hard thing with these elements (faces, edges and nodes) is that the information on them may be incomplete. For example, a face in the ghost layer may only have an element on one side because the other element is not loaded on the process.

kernel/Utilities/SparsityEstimator.cpp
```diff
@@ -15,310 +18,416 @@ SparsityEstimator::SparsityEstimator(const GlobalIndexing& rowindexing,
...
 
-    /// Get a reference to either the owned or notOwned DoF set.
-    std::map<std::size_t, std::size_t>& getDoFs(bool owning) {
-        return owning ? owned : notOwned;
-    }
+/// Set of DoFIntervals
+using DoFIntervals = std::set<DoFInterval>;
+
+/// Enum to specify rows or columns of a matrix
+enum MatrixSide { ROWS, COLUMNS };
```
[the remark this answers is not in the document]
Laar: Hmm, that would give me the impression that we are talking about how something is ordered. The word side is not really nice, because that could also mean left or right hand side of the matrix.

Here I'm just wanting to know whether a structure is describing the rows or the columns of a matrix. Thinking of it `RowsOrColumns` sounds a bit long/strange. So have to think about it.

kernel/Utilities/SparsityEstimator.cpp
Laar: Replaced

kernel/Utilities/SparsityEstimator.cpp
Laar: Maybe, it feels a bit strange to have a begin and end on a 2D table. I'll have a thought about it.

kernel/Utilities/SparsityEstimator.cpp
Laar: Could but with a horrible complexity. As it would currently need to traverse all faces of the mesh.

I think it may be more useful to have a function to check if a face is a subdomain boundary. I'll add that.

kernel/Utilities/SparsityEstimator.cpp
Laar: Replaced.

Laar:
> The sparsity estimator is simpler now. I am still struggling with the general structurre of the code. This is still a nice simplification. On some level it would be nice to make Table2D a specialisation of MiddleSizeMatrix, but then we would have to template that.

I would actually think the reverse, MiddleSizeMatrix is a specialization of a 2D table. One where you are restricted to numbers and gain a lot of matrix operations. However, I did not plan to do that to allow easier replacement of the matrix/vector classes by a library.

kernel/Utilities/GlobalMatrix.h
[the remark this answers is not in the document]
JensWehner: yeah I have no good idea either. Then leave it as it is.

kernel/Utilities/SparsityEstimator.cpp
JensWehner: Ah okay. Thanks.

Laar:
> I am still struggling with the general structurre of the code.

Anywhere specific that could use some care/explanation?

kernel/Utilities/GlobalMatrix.cpp
```diff
@@ -166,7 +167,7 @@ GlobalPetscMatrix::GlobalPetscMatrix(const GlobalIndexing& rowIndexing,
     // Dummy call to always have an valid matrix in A_.
     MatCreateSeqAIJ(PETSC_COMM_SELF, 0, 0, 0, PETSC_NULL, &A_);
 
-    reinit();
+    reinit(false);
```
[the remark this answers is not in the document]
Laar: As in a member variable of the class?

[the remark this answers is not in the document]
JensWehner: no I mean just 
```suggestion
  bool uselements=false;
    reinit(uselements);
```

with a better name than uselements

kernel/Utilities/SparsityEstimator.cpp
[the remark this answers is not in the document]
JensWehner: `Indexing` ?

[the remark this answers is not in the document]
Laar: Indexing is already used for something else (for each row/column to which degree of freedom does it correspond).

I've changed it to `RowsOrColumns` find that rather direct and it does not take that much more space.

kernel/Utilities/GlobalMatrix.cpp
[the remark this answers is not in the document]
Laar: Done

JensWehner (approved):

Laar:
I'll see that I fix the problem with the mac os run at some point.

JensWehner (approved):

JensWehner:
@Laar do you have any idea why the tests fail? Old petsc version?

Laar:
> @Laar do you have any idea why the tests fail? Old petsc version?

I think we need to add some external solver package. I changed that we need one for  `010GlobalAssembly_SelfTest.cpp ` and `040Maxwell_SelfTest.cpp `. But apparently it is not installed by default on mac.

PR #67, opened 2020-12-12
Cmake object libraries

Laar:
First step in further improving the build system. This replaces the separate libraries for each of the kernel-components by cmake object-libraries. The advantage is that the separate components don't create extra shared/static libraries, instead there is a single final hpgem-library.

Note this was not done earlier as linking object libraries through `target_link_libraries([target] [object-lib])` was only introduced in cmake 3.9, while the previous minimum required cmake version was 3.3.


JensWehner (approved):
Really nice change.

PR #68, opened 2020-12-12
Break BasisFunctionSet QuadratureRule dependency loop

Laar:
The BasisFunctionSet depended on QuadratureRule and vice versa. This dependency was introduced when adding caching for basisfunction value at quadrature points. Ideally the design back then would have been to build this on top of  `BasisFunctionSet` and `GaussQuadratureRule`, so that neither of these two needs to know of the other.

However, a different design was chosen where `GaussQuadratureRule` stores the evaluations of the basis functions in the `BasisFunctionSet`. This added a dependency of `GuassQuadratureRule` on `BasisFunctionSet` for the actual evaluation. The reverse dependency was two fold. 
 1. When the BasisFunctionSet disappears the evaluations in `GaussQuadratureRule` need to be cleared to prevent memory leaking. This requires a notifcation from `BasisFunctionSet` to the quadrature rule, introducing a dependency.
 2. To evaluate the basis functions at quadrature points there were a bunch of forwarding functions from `BasisFunctionSet` to `GaussQuadratureRule`.

This PR removes these latter two dependencies. 

The first one is removed by registering a callback function instead of a specific quadrature rule. This is not the most elegant situation, but in my opinion the best solution to this inelegant problem. 

The second one is removed by inlining/removing the functions from `BasisFunctionSet`.  This is a bit verbose, as `evalCurl` is now templated to the dimension instead of having `eval2D` and `eval3D`. The resulting template specialization required moving the implementation to the header.


JensWehner (approved):
a) Your pull request explanations are very helpful. Thanks

b) Yeah this solution is not brilliant either, but it is certainly an improvement on the old design.

Laar:
Thanks for reviewing even now.

PR #70, opened 2020-12-13
FE namespace (following #69)

Laar:
Following #69, introduce the FE namespace.


JensWehner (approved):
ahh I see

Laar:
> ahh I see

Note that #69 added one thing that is not obvious from this PR. By introducing an Object Library you have to specify its compile dependencies. This may (not sure) enforce some restrictions that follow from the limited dependencies.

Issue #61, opened 2020-12-02
Add metadata to elements


Laar:
The current mesh is purely a mesh without any meta-data on the elements. This is a problem when there are properties that vary depending on the mesh location, for example material properties. At the moment these need to be hardcoded, which is both error prone and non-scalable.

As solution to this problem we need to add some form of metadata to the elements (and possibly also to the faces) of the mesh. To do this we need to add the metadata in several locations:
 1. The preprocessor datastructures
 2. The mesh reader(s) of the preprocessor (e.g. the one for centaur meshes)
 3. The dataformat used to transfer the mesh between the preprocessor and the kernel needs to be ammended
 4. The preprocessor mesh writer
 5. The reader in the kernel needs to read it

Laar:
Done

PR #71, opened 2020-12-14
Add more flexible definition of the IW crystal

Laar:
Temporary work till #61 is fixed, a new definition of the inverse woodpile similar to that of structure 4.

I did not replace structure 4 as that can be kept to match the original meshes with a slightly different design.


applications/DG-Max/ElementInfos.cpp
```diff
@@ -159,6 +200,26 @@ double jelmerStructure(const Geometry::PointPhysical<DIM>& pPhys,
         // std::cout << pPhys[0] << " " << pPhys[1] << " " << pPhys[2]  << "
         // " << epsilon_ << "\n";
 
+    } else if (structureType == 5) {
```
JensWehner, line 203: can we at least make the structure Type an enum?

JensWehner (approved):
LGTM

applications/DG-Max/ElementInfos.cpp
Laar: Should have known this comment was coming, I'll see what I can do. Problem is that the input goes via a command line argument.

Laar: Done

PR #73, opened 2020-12-16
 Centaur reader improvements

Laar:
Improvements in the Preprocessor as first part of #61 

 - Read zone information from centaur meshes
 - Documentation on the centaur mesh format
 - Introduce `MeshSource` to clarify the interface the mesh reader expects from the mesh

TODO:

 - [x] Add zone info to `MeshSource`
 - [x] Add zone info to the preprocessor mesh


applications/Preprocessor/MeshSource.h
```diff
@@ -0,0 +1,113 @@
...
+#define HPGEM_MESHSOURCE_H
+
+#include <vector>
+#include "customIterator.h"
+
+namespace Preprocessor {
+
+/**
+ * Source of a mesh
+ */
```
JensWehner, line 48: This explanation could be a bit more verbose. :)

Laar: It should be a bit better now

JensWehner: It is much better

JensWehner (approved):
LGTM

PR #76, opened 2020-12-21
Remove QHull

Laar:
QHull was used to implement an unstructured mesh generator in the
kernel. However, this mesh generator has not been working (or even
compiling) for a long time. The Qhull code would need to be ported to
the Preprocessor for it to become viable again. Moreover, in the current
user/developer base there does not seem to be a desire for using this
custom mesh generator.

Thus at the moment the QHull dependency and build only cause anoying CI
errors and does not provide a useful feature. Hence this commit removes
it. (If desired it can be readded in a correct fashion later on.)


JensWehner (approved):
ohh yes. :)

PR #77, opened 2021-01-04
DGMaxEigenvalue workspace inner class update

Laar:
Similar to a change done in #75, this updates `DGMaxEigenvalue` to also use inner classes for its implementation details (similar to `DivDGMaxEigenvalue` in #75). This should prevent accidental name clashes in the future.

Note:
 - This is mainly a mechanical update, no details were considered (nor changed)
 - The structure of DGMaxEigenvalue and DivDGMaxEigenvalue is very similar. Removing this duplication would be nice, but both algorithms are currently in such a state of flux that I don't think it is advisable to do so now.


JensWehner (approved):
LGTM.

PR #78, opened 2021-01-04
Remove Global->Local Vector taking real part

Laar:
Taking a global vector and writing it as local vector (time integration one) did take the real part of the coefficients. This is strange, as
this discards any imaginary part of the solution. This is especially problematic when plotting the solution or locally computing an error, as there the imaginary part will have been discarded.

Search through the history I could not find a reason for taking the real part (though I did remove it unintentionally and undid that at r925-926).


JensWehner (approved):
Yes that is really weird.

PR #79, opened 2021-01-12
Improve vtk writer

Laar:
Preparatory improvements of the `VTKSpecificTimeWriter`, which is used to write the FEM solution for later viewing in paraview.

The basic idea of this writer is that it outputs 1 master `pvtu` file, which directs to one `.vtu` file for each processor in an MPI computation. The processor with rank 0 writes the master file, and each processor writes its own `vtu` file with the data for the local part of the mesh. The usual sequence (both before and afterwards) is something like
```
VTKSpecificTimeWriter writer ("file-prefix"); // Writes the start of the files and the local meshes
writer.write([result computing function], "name of data"); // Writes a single scalar/vector/tensor value to the output
....
[implicit] ~writer(); // Adds all closing tags and closes the files
```

This PR is the first in two parts in supporting higher order elements, and is mainly preparatory work. The main parts are:
 1.  The VTKSpecificTimeWriter had a very long constructor writing both the local and master files. Writing these files has been split of into their own functions.
 2.  The writing of base64 encoded binary data is now delegated to a function
 3.  For the local mesh part, the data is now first computed and then written all to file, greatly improving the readability.
 4.  Added testing, based on manually verified correct outputs.

Reviewing notes:
 - A large part of the `VTKSpecificTimeWriter` constructor was moved and partially rewritten and reformatted. I don't think the diffview here is practical for viewing.
 - The files in `tests/files` are just some simple meshes for testing
 - The files in `tests/self/Output/vtk-reference` were manually checked by loading them in paraview and comparing them to the expected output.


JensWehner (approved):
Looking good. The cmake comparison of files is nice

Laar:
> Looking good. The cmake comparison of files is nice

Thanks

PR #80, opened 2021-01-12
VTK writer: Lagrange elements (after #79)

Laar:
Part two of updating the VTKSpecificTimeWriter. Recommendation to first merge #79, or at least look at the differences with respect to that PR.

This adds support for outputting the arbitrary order Lagrange elements from VTK, as described [here](https://blog.kitware.com/modeling-arbitrary-order-lagrange-finite-elements-in-the-visualization-toolkit/).

For VTK/paraview this allows representing both curvilinear elements (i.e. non straight edges/faces) and higher order polynomial solutions (i.e. not linear). HPGEM does not have curvilinear elements, but we do use higher order polynomials for approximating the solution, with this change we can correctly visualize these. As example see 
image
which is a rendering of a quadratic function on a SINGLE cube (output `unitCube-quadratic.pvtu`).

To support this we needed two parts:
 1. Decouple the one-to-one mapping of corner points to points in the vtk-output, as the number of points needed to describe a Lagrange element depends on the order.
 2.  A way of generating the Lagrange points for each shape for each order (linear, quadratic, etc.) in the ordering that vtk/paraview expects them.

Part one is done by introducing the `VTKElement` which describes the types of cells/elements that can be outputted to the vtk files. For vtk reference geometries (line, triangle, square, etc.) one derives a subclass, and registers it in the `VTKSpecificTimeWriter`. While we could completely switch over to the Lagrange elements, we kept the original linear shapes  in `VTKStandardElements`, as these are far simpler and may be useful for compatibility.

Part two is the far more complicated part, but mainly due to the complicated order in which VTK expects the points for some shapes and the lack of documentation. When the wrong ordering is used the plot will be completely deformed. The general approach is to list the points as follows:
 - First the corner points of the shape
 - Then the points on the edges of the shape, usually starting at the first corner and going in a counter clockwise direction
 - For 3D shapes the points on the faces
 - The points in the interior

For several shapes this is straightforward (e.g. the line/curve), but for a few shapes it is more complicated:
 - The ordering of the 12 edges of a cube is strange.
 - The triangle and tetrahedron use a recursive approach, the interior points (if any) are again a triangle/tetrahedron for which the process is repeated
 - To list the points on the triangular faces of the tetrahedron the algorithm for the triangle is used (minus the points on the outer edge), making it doubly complicated.


kernel/Output/VTKLagrangeCurve.h
```diff
@@ -0,0 +1,62 @@
...
+class VTKLagrangeCurve : public VTKElement<1> {
+   public:
+    explicit VTKLagrangeCurve(std::size_t order);
+
+    std::uint8_t vtkId() const final { return 68; }
+
+    const std::vector<Geometry::PointReference<1>>& getPoints() const final {
+        return points_;
+    }
+
```
JensWehner, line 55: private
```suggestion
private:
```

kernel/Output/VTKLagrangeHexahedron.h
```diff
@@ -0,0 +1,60 @@
...
+class VTKLagrangeHexahedron : public VTKElement<3> {
+   public:
+    VTKLagrangeHexahedron(std::size_t order);
+
+    std::uint8_t vtkId() const final { return 72; }
+
+    const std::vector<Geometry::PointReference<3>>& getPoints() const final {
+        return points_;
+    }
+
```
JensWehner, line 54:
```suggestion
private:
```

kernel/Output/VTKLagrangeQuadrilateral.h
```diff
@@ -0,0 +1,59 @@
...
+class VTKLagrangeQuadrilateral : public VTKElement<2> {
+   public:
+    explicit VTKLagrangeQuadrilateral(std::size_t order);
+
+    std::uint8_t vtkId() const final { return 70; }
+
+    const std::vector<Geometry::PointReference<2>>& getPoints() const final {
+        return points_;
+    }
+
```
JensWehner, line 53:
```suggestion
private:
```

kernel/Output/VTKLagrangeTetrahedron.h
```diff
@@ -0,0 +1,59 @@
...
+namespace Output {
+class VTKLagrangeTetrahedron : public VTKElement<3> {
+   public:
+    explicit VTKLagrangeTetrahedron(std::size_t order);
+    std::uint8_t vtkId() const final { return 71; }
+
+    const std::vector<Geometry::PointReference<3>>& getPoints() const final {
+        return points_;
+    }
+
```
JensWehner, line 53:
```suggestion
private:
```

kernel/Output/VTKLagrangeTriangle.cpp
```diff
@@ -0,0 +1,107 @@
...
+        points_[i][1] = h * bary[i][2];
+    }
+}
+
+std::vector<std::valarray<std::size_t>>
+    VTKLagrangeTriangle::computeBaryIntegerPoints(std::size_t order) {
+
+    // Integer variation on barycenteric coordinates. When multiplying by h =
+    // 1/order these are actual barycentric coordinates. While the actual
+    // coordiantes of the points don't matter, they will be mapped such that
```
JensWehner, line 62:
```suggestion
    // coordinates of the points don't matter, they will be mapped such that
```

kernel/Output/VTKLagrangeTriangle.cpp
```diff
@@ -0,0 +1,107 @@
...
+    std::vector<IVec> points(numPoints);
+
+    std::size_t index = 0;  // Index in the output
+    for (std::size_t layer = 0; 3 * layer <= order; ++layer) {
+        // Split the points on this layer into offset + localPoint
+        // Offset is a vector that ensures that we are away from each corner
+        // point by 'layer'-points. The localPoint is a localPoint on the layer,
+        // which is ensured by having at least one zero coordinate.
+        IVec offset = IVec({layer, layer, layer});
+        // sum(offset[i]) = 3*layer, so ew need sum(localPoint[i]) = rOrder
```
JensWehner, line 76:
```suggestion
        // sum(offset[i]) = 3*layer, so we need sum(localPoint[i]) = rOrder
```

kernel/Output/VTKLagrangeTriangle.h
```diff
@@ -0,0 +1,84 @@
...
+class VTKLagrangeTriangle : public VTKElement<2> {
+   public:
+    explicit VTKLagrangeTriangle(std::size_t order);
+
+    std::uint8_t vtkId() const final { return 69; }
+
+    const std::vector<Geometry::PointReference<2>>& getPoints() const final {
+        return points_;
+    }
+
```
JensWehner, line 56:
```suggestion
private:
```

kernel/Output/VTKSpecificTimeWriter_Impl.h
```diff
@@ -447,5 +435,62 @@ void VTKSpecificTimeWriter<DIM>::writePaddedTensor(
...
+template <std::size_t DIM>
+void VTKSpecificTimeWriter<DIM>::setupMapping(std::size_t order) {
+    logger(ERROR, "No VTK element mapping in dimension %", DIM);
+    // Specialized per DIM
+}
+
+template <>
+inline void VTKSpecificTimeWriter<0>::setupMapping(std::size_t order) {
+    elementMapping_[Geometry::ReferenceGeometryType::POINT] =
+        std::shared_ptr<VTKElement<0>>(new VTKPoint());
```
JensWehner, line 447:
```suggestion
    elementMapping_[Geometry::ReferenceGeometryType::POINT] =
        std::make_shared<VTKPoint>();
```

kernel/Output/VTKSpecificTimeWriter_Impl.h
```diff
@@ -447,5 +435,62 @@ void VTKSpecificTimeWriter<DIM>::writePaddedTensor(
...
+}
+
+template <>
+inline void VTKSpecificTimeWriter<1>::setupMapping(std::size_t order) {
+    if (order == 1) {
+        elementMapping_[Geometry::ReferenceGeometryType::LINE] =
+            std::shared_ptr<VTKElement<1>>(new VTKLine());
+    } else {
+        elementMapping_[Geometry::ReferenceGeometryType::LINE] =
+            std::shared_ptr<VTKElement<1>>(new VTKLagrangeCurve(order));
```
JensWehner, line 457: see above

kernel/Output/VTKStandardElements.h
```diff
@@ -0,0 +1,138 @@
...
+#define HPGEM_VTKSTANDARDELEMENTS_H
+
+#include "VTKElement.h"
+
+namespace hpgem {
+namespace Output {
+
+/// Simple hardcoded (linear) elements
+
+class VTKPoint : public VTKElement<0> {
```
JensWehner, line 48:
```suggestion
class VTKPoint final: public VTKElement<0> {
```

you can declare the whole class final if you want.

JensWehner (approved):
Small nitpicks otherwise fine

Laar:
Thanks for the suggestions (just implemented them).

PR #83, opened 2021-01-14
Centaur reader fixes

Laar:
Basic fixes for reading 2D centaur meshes. Which I apparently did not sufficiently test enough.

Reading of boundary information is at the moment useless (and discarded), but it provided good insight into whether the reading process was good or not. As it is far easier to check if a name makes sense, than some seemingly random numbers. 

Because there may be use for this boundary information, either for future debugging, or actually in the preprocessor, I have left the reading code in.

### DIFF
--- a/applications/Preprocessor/centaur.cpp
+++ b/applications/Preprocessor/centaur.cpp
@@ -78,12 +78,21 @@ namespace Preprocessor {
  *    capability. The number of lines is such that the first Nl-1 lines contain
  *    exactly Ne entries, and the last line contains the remaining entries.
  *
+ *    This is the most common format. For some things (zones, periodic boundary
+ *    conditions) the data does not follow this format. For example, Zones in 3D
+ *    interleave the zone-element-indices and the zone names (instead of having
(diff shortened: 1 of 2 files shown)


applications/Preprocessor/centaur.cpp
```diff
@@ -237,7 +233,9 @@ CentaurReader::CentaurReader(std::string filename) {
     // Interface panels (3D only)
     // 1 data line with number N of interface panels (could be zero)
     // 1 data line with N entries (panel numbers?)
-    skipGroup();
+    if (is3D()) {
+        skipGroup();
+    }
```
JensWehner, line 238: The comment and the command together are a bit misleading

JensWehner (approved):

applications/Preprocessor/centaur.cpp
Laar: You mean that there are two data lines listed?

Because that is the standard 'group' format (see also point 4 at the top of the file).

PR #85, opened 2021-01-18
Update codecov.yml

JensWehner:
Let us see if it has a noticeable effect.


JensWehner:
@Laar it is like 15 min now.

Laar:
Looks good to me. Interesting though that there is such a big impact. I guess from inlining a lot of methods.

Laar (approved):

PR #87, opened 2021-01-19
Gmsh-Reader

JensWehner:
Adding Gmsh-Reader


applications/Preprocessor/CMakeLists.txt
```diff
@@ -1,44 +1,46 @@
...
+	)
+	target_link_libraries(Preprocessor-lib
+		PUBLIC
+            HPGEM::HPGEM)
+            
 
 if(hpGEM_USE_MPI)
-    target_link_libraries(Preprocessor.out MPI::MPI_CXX)
-    target_link_libraries(ExtractSurface.out MPI::MPI_CXX)
+    target_link_libraries(Preprocessor-lib PUBLIC MPI::MPI_CXX)
```
Laar, line 18: It might be more practical if we fix this higher up, for example make it a public dependency in `kernel/Base`.

applications/Preprocessor/CMakeLists.txt
```diff
@@ -1,44 +1,46 @@
...
 if(hpGEM_USE_SLEPC)
-    target_link_libraries(Preprocessor.out SLEPc::SLEPc)
-    target_link_libraries(ExtractSurface.out SLEPc::SLEPc)
+    target_link_libraries(Preprocessor-lib PUBLIC SLEPc::SLEPc)
 endif()
+
+#the preprocessor does not perform floating poing arithmetic, so it doesn't need special Petsc/not petsc guards
+add_executable(Preprocessor.out
+               main.cpp
+               ../../kernel/Logger.cpp
```
Laar, line 36: If you switch to linking to hpgem, why then still include these source files?

applications/Preprocessor/CMakeLists.txt
```diff
@@ -1,44 +1,46 @@
...
-add_executable(ExtractSurface.out
-               extractSurface.cpp
-               centaur.cpp
-               hpgem.cpp
-               ../../kernel/Logger.cpp
-               ../../kernel/Base/CommandLineOptions.cpp
-               ../../kernel/Base/MpiContainer.cpp)
-
-set(CURRENT_TARGET Preprocessor.out)
-include(${CMAKE_SOURCE_DIR}/conf/cmake/CLionFix.cmake)
```
Laar, line 21: Don't we need this anymore?

applications/Preprocessor/lib/gmsh.h
```diff
@@ -0,0 +1,82 @@
...
+#define HPGEM_APP_GMSH_H
+
+#include "MeshSource.h"
+#include <array>
+#include <vector>
+#include <fstream>
+#include <string>
+#include <map>
+
+// Most of this was taken from https://github.com/Applied-Scientific-Research/gmsh-reader/tree/main/src
```
Laar, line 49: Depending on what is taken, this will be a License issue, as that code is GPL and this codebase is BSD3.

tests/unit/Preprocessor/007Gmsh_UnitTest.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include "Logger.h"
+
+#include "../catch.hpp"
+
+using namespace hpgem;
+TEST_CASE("010Gmsh_UnitTest", "010Gmsh_UnitTest") {
+
+    std::string filename =
+        std::string(PREPROCESSOR_TEST_DATA_FOLDER) + "/insulated-4.1.msh";
+    Preprocessor::GmshReader test(filename);
```
Laar, line 55: Looks like this is more a SelfTest compared to a unit test

Laar (commented):
Some initial feedback. Restructuring of the preprocessor seems like a good idea.

applications/Preprocessor/CMakeLists.txt
JensWehner: I do not know actually, I do not use CLion. :)

applications/Preprocessor/CMakeLists.txt
JensWehner: I do not link to HPGEM, only to a small part of the kernel files. It was sort of like this before, so I figured I keep it. BUt I do not care either way.

applications/Preprocessor/lib/gmsh.h
JensWehner: yeah, actually good point. I should know. I delete the stuff.

tests/unit/Preprocessor/007Gmsh_UnitTest.cpp
JensWehner: What is a self-test?

My reasoning was as follows: a) it is a class so it is a unit_test. b) I wanted to use the catch2 stuff. 
Also, there was no tests for the preprocessor so far, so I thought. 
I can put it somewhere else, what do you prefer?

applications/Preprocessor/lib/gmsh.h
JensWehner: The getline stuff is stackoverflow, so that is fine.

JensWehner: The locate_in_file part is borderline, because that is pretty generic. The rest will go away as our datastructures are different anyway.

tests/unit/Preprocessor/007Gmsh_UnitTest.cpp
Laar: Local name for integration tests.

My basic expectation of a unit test is that it should be fully deterministic and should be contained in the system. So no things like IO. But the question is if and how strict we actually adhere to this.

applications/Preprocessor/lib/gmsh.h
Laar: Stackoverflow seems to be CC-BY-SA of some version (depending on the date). So depending on how inventive it is we might need to update things (but this seemed fine).

applications/Preprocessor/CMakeLists.txt
Laar: I'll try then.

applications/Preprocessor/CMakeLists.txt
Laar: I think you do in lines 12-14. So we can link either to the whole of hpGEM (seems a bit overkill, but easier), or just to the essential functions. But doing both seems to be a bit double.

JensWehner: Ahh you are right, it is public. Good point. I think just linking to the whole of hpGEM is the way to go.

applications/Preprocessor/CMakeLists.txt
JensWehner: Ahh you use CLion?

applications/Preprocessor/lib/gmsh.h
JensWehner: yes, to my knowledge posting the link to the thread is enough to fulfill the license agreement

Laar (commented):
Looks like `customVector.h` is not used, so maybe remove that instead of moving it.

applications/Preprocessor/lib/gmsh.h
```diff
@@ -0,0 +1,85 @@
...
+
+#include "MeshSource.h"
+#include <array>
+#include <vector>
+#include <fstream>
+#include <string>
+#include <map>
+
+// Most of this was taken from
+// https://github.com/Applied-Scientific-Research/gmsh-reader/tree/main/src
```
Laar, line 50: Did you check the license for this?

Laar (commented):
Looks good. A few minor things to fix at some point:
 - Naming of functions, some functions start with a capital letter, some not.
 - Some documentation and comments in between would be nice.

applications/Preprocessor/lib/gmsh.h
JensWehner: I found the stackoverflow post for the part I am using, so this is fixed now.

JensWehner:
Actually I am going to close this PR, because we had to switch to the 2.2. version because mesh.io

PR #88, opened 2021-01-25
Divdgmax slepc shift invert

Laar:
Background:

For DivDGMax we want to solve an eigenvalue problem of the form 
```Sx = ω^2Mx```
 for the lowest (positive) ω^2 and thus the smallest positive eigenvalues. Assuming that 0 is not an eigenvalue this is equivalent to solving 
```ω^(-2) Sx = Mx```
which then looks for the largest eigenvalues `ω^(-2)`. This latter approach is favourable from a numerical point of view, as the eigenvalues at the end of the spectrum converge first, which in the second case are exactly the wanted eigenvalues.

Change:
The old code directly implemented the second form. The update changes to using the first form, and letting the Spectral Transform (ST) functionality of Slepc internally transform it to using the second form. The advantage of this is threefold:
 1. The more interesting eigenvalues `ω^2` are returned
 2. The eigenvectors are normalized with respect to `M`, which is better than with respect to `S`.
 3. The eigenvalue problem is now Generalized Hermitian according, which could allow a more efficient solve and should ensure that the eigenvalues are purely real.

The PR is in draft till I can test whether 3. actually gives some benefit.

### CODE REVIEW

File: applications/DG-Max/Utils/Symmetry/Domain.h [hunk 1]
```diff
@@ -0,0 +1,132 @@
+/*
```
JensWehner, line 1: How does this fit in?

applications/DG-Max/Algorithms/DivDGMaxEigenvalue.cpp
```diff
@@ -50,6 +50,8 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 using namespace hpgem;
 
+// Helper function
+
```
JensWehner, line 54: ?

applications/DG-Max/Algorithms/DivDGMaxEigenvalue.cpp
```diff
@@ -185,15 +187,27 @@ class DivDGMaxEigenvalue<DIM>::SolverWorkspace {
...
-        // a spectral transformation.
-        // It may be possible to remove this relic with the current version of
-        // SLEPc.
-        err = EPSSetOperators(solver_, massMatrix_, stiffnessMatrix_);
+        err = EPSSetOperators(solver_, stiffnessMatrix_, massMatrix_);
+        CHKERRABORT(PETSC_COMM_WORLD, err);
+
+        // The mass matrix is positive semidefinite, the stiffness matrix is
+        // Hermitian indefinite (k-phase-shifts are also Hermitian for real k).
+        err = EPSSetProblemType(solver_, EPS_GHEP);
```
JensWehner, line 195: do we have to bump the Slepc version in cmake?

JensWehner (commented):
This makes it more clear, the tesselation argument class seems to not belong into this PR.

applications/DG-Max/Utils/Symmetry/Domain.h
Laar: Crap, that should not have been committed.

applications/DG-Max/Algorithms/DivDGMaxEigenvalue.cpp
Laar: No, all the needed functionality is long there in SLEPc. But probably nobody knew about it or something like that.

Laar:
Fixed the issues. The GHEP does not seem to affect the performance much, so leaving that in.

JensWehner (approved):

PR #91, opened 2021-02-09
Remove deprectated Physical[Geometry] classes

Laar:
These are no longer used and deprecated in favor of the more general PhysicalGeometry class.


Laar:
Needs more work to fix the unit tests

Laar:
@JensWehner Finally got around to fixing the tests (was simpler that expected), so this is ready for review. I think you'll like this one.

JensWehner (approved):
Sorry least important one, but that was quick.

PR #92, opened 2021-02-22
Update .codecov.yml

JensWehner:


Laar (approved):
Don't know what it exactly does. But I assume that it is ok.

JensWehner:
It removes these codecov comments from the middle of the PR when doing code review.

PR #94, opened 2021-02-23
Cmake improvements

Laar:
Two improvements to the cmake configuration:
 1. Expose git hash, ref and dirty state (and include it in the DGMax output)
 2. Adapt the slepc find module to allow it to find slepc.pc (in addition to SLEPc.pc)


JensWehner (approved):
Nice changes, thanks. :)

Laar:
I know that the code in SavageHutter is ugly. However, I wanted to spend the least amount of maintenance on a code that is probably not used any more.

PR #95, opened 2021-02-24
Cleanup namespaces and includes in the Preprocessor

Laar:
This moves the `using namespace hpgem` into the Preprocessor namespace
scope. For that it was required to introduce this namespace explicitly
in some of the cpp files.

Additionally, some unused include statements were removed.


Laar:
Oops, I realize now that I also committed the slight restructuring of the preprocessor main file. That should have been split in a different commit. The idea is to pull out the partitioning into a separate method, as I soon need an additional processing step.

JensWehner (approved):

PR #98, opened 2021-02-28
Field overlap in DGMax

Laar:
Adds computing overlap integrals between the modes for one wave-vector (=k-point) and those of the next wave-vector. For each pair of wave-vectors this would result in a N by M matrix, where N and M are the number of modes at those wave-vectors. The expectation is that this may provide sufficient information to correctly connect the bands.

The computation of the integrals is the simplest part. It is just computing the product  `v_i^H M v_j`, where M is the mass matrix and v_i and v_j are the modes at the two wave-vectors. The more complicated part is that we need to both store the current and previous set of eigenvectors. To simplify the required juggling with the eigenpairs, I introduced the `Eigenpairs` class. A significant part of the PR is replacing the previous storage by the Eigenpairs.

Reviewing advice:
 - Look at the Eigenpairs class first, the replacements in the other classes will make much more sense.
 - The method to implement by the solvers is introduced in `AbstractEigenvalueResult.h`
 - After that look at the rest


applications/DG-Max/Algorithms/DGMaxEigenvalue.cpp
```diff
@@ -671,29 +635,24 @@ void DGMaxEigenvalue<DIM>::SolverWorkspace::initStiffnessMatrixShifts() {
...
+    eigenpairs_.loadEigenpairs(solver_, tempFieldVector_);
     CHKERRABORT(PETSC_COMM_WORLD, error);
+    // Reorder
+    std::vector<std::size_t> ordering(eigenpairs_.size());
+    std::iota(ordering.begin(), ordering.end(), 0);
+    std::sort(ordering.begin(), ordering.end(),
+              [&](const std::size_t& i1, const std::size_t& i2) {
+                  PetscScalar e1 = eigenpairs_.getEigenvalue(ordering[i1]);
+                  PetscScalar e2 = eigenpairs_.getEigenvalue(ordering[i2]);
+                  if (PetscRealPart(e1) != PetscRealPart(e2)) {
```
JensWehner, line 649: this line still worries me a bit, because of floating point, but I guess if petsc solves for complex pairs, the real parts are exactly equal.

Laar: I don't see any problem with it. The goal is to provide an order on the complex numbers so that the result will be sorted and preferably in a reproducible fashion. This just defines a total order that orders the values by real part, and to make it a total order it sorts it based on imaginary as backup.

applications/DG-Max/Algorithms/DivDGMaxEigenvalue.cpp
```diff
@@ -123,31 +120,15 @@ class DivDGMaxEigenvalue<DIM>::SolverWorkspace {
         PetscInt iterations;
         err = EPSGetIterationNumber(solver_, &iterations);
         CHKERRABORT(PETSC_COMM_WORLD, err);
-        err = EPSGetConverged(solver_, &numberOfConvergedEigenpairs);
-        CHKERRABORT(PETSC_COMM_WORLD, err);
+        std::swap(eigenpairs_, previousEigenpairs_);
```
JensWehner, line 123: clever I like this

applications/DG-Max/Algorithms/DivDGMaxEigenvalue.cpp
```diff
@@ -157,14 +138,35 @@ class DivDGMaxEigenvalue<DIM>::SolverWorkspace {
...
         tempVector_.writeTimeIntegrationVector(timeIntegrationVector);
     }
 
+    LinearAlgebra::MiddleSizeMatrix computeOverlapIntegrals() {
+        LinearAlgebra::MiddleSizeMatrix result(eigenpairs_.size(),
+                                               previousEigenpairs_.size());
+        PetscErrorCode err;
+        for (std::size_t i = 0; i < eigenpairs_.size(); ++i) {
+            err = MatMult(massMatrix_, eigenpairs_.getEigenvector(i),
+                          tempVector_);
```
JensWehner, line 155: Is the member tempVector_ necessary for performance? otherwise this method could be `const`

applications/DG-Max/Algorithms/DivDGMaxEigenvalue.cpp
```diff
@@ -245,10 +240,11 @@ class DivDGMaxEigenvalue<DIM>::Result final
            const Base::MeshManipulator<DIM>* mesh,
            const DivDGMaxDiscretization<DIM>& discretization)
         : workspace_(workspace), mesh_(mesh), discretization_(discretization) {
-        auto& eigenvalues = workspace.getEigenvalues();
-        frequencies_.resize(eigenvalues.size());
+        auto& eigenpairs = workspace.getEigenpairs();
```
JensWehner, line 243:
```suggestion
        const auto& eigenpairs = workspace.getEigenpairs();
```

kernel/Utilities/Eigenpairs.cpp
```diff
@@ -0,0 +1,131 @@
...
+    err = VecDuplicateVecs(sample, newSize, &eigenvectors_);
+    CHKERRABORT(PETSC_COMM_WORLD, err);
+    numberOfEigenvectors_ = newSize;
+}
+
+/**
+ * \brief Load eigenpairs from a solver
+ *
+ * @param eps The solver to load the eigenpairs from
+ * @param sample A sample vector to duplicate for storing eigenvectors
```
JensWehner, line 94: I do not understand the idea behind sample, it is a vector with the right shape, which is not used?

JensWehner (approved):
a) Thank you for the review advice that is a really nice practice
b) Good work, I like the storage Eigenvalues
c) The isolated classes maybe could benefit from some unittests
d) The abstracteigensolver class gets more and more methods which link it to more concrete Problems. Maybe a renaming could be suitable soon. 

Thanks good work.

kernel/Utilities/Eigenpairs.cpp
Laar: It is a vector which has the correct type (e.g. not a block vector) and layout (the number of entries per processor). The content of the vector is not used, but it is used as a cloning template. By cloning it we get the actual storage for the eigenvectors.

applications/DG-Max/Algorithms/DivDGMaxEigenvalue.cpp
Laar: It is needed because `MatMult(M, x, y)`  corresponds to `y = Mx` has the requirement that x and y must not be the same vector. Which I guess is indeed for performance reasons.

What I could do is make `tempVector_` mutable, which sounds quite reasonable as it is just a vector for temporary storage.

Laar:
> a) Thank you for the review advice that is a really nice practice
> b) Good work, I like the storage Eigenvalues
> c) The isolated classes maybe could benefit from some unittests
> d) The abstracteigensolver class gets more and more methods which link it to more concrete Problems. Maybe a renaming could be suitable soon.
> 
> Thanks good work.

a) Will keep it up then
c) Eigenpairs might indeed do, but it is quite a work to set up. As it requires setting up a dummy eigenvalue problem and solving it. Might be doable, but have to think about how to do it efficiently
d) Abstract is more meant here as a base class with little or no implementation, where the deriving class fills in the holes. It is indeed getting more and more specialized methods, but the specialization is happening within the context of the DGMax namespace (or should be at some point), for which all these methods make sense.

applications/DG-Max/Algorithms/DivDGMaxEigenvalue.cpp
Laar: Looking further into it. Unfortunately all the PETSc and const don't really work well together. All the types (Mat, Vec, etc) are typdef's to pointer types (i.e. `typedef Vec somestruct*`). So making them all const (e.g. `const Vec`) would only make the pointer const, not the target matrix, vector, etc. So I don't see a practical way to make it more const safe.

PR #99, opened 2021-03-01
Rewrite centaur reader

Laar:
Rewrite the implementation of the centaur reader with as primary goals simplifying it and making it more robust.

The primary differences are:
 1. It reads in all the nodes and elements at the start, in stead of on-demand. This is less memory efficient but removes a whole lot of complexity.
 2. The 'unformattedFile' is replaced by `FortranUnformattedFile`, a reader for the standard way Fortran does the unformatted output. Following [the advice from gcc](https://gcc.gnu.org/onlinedocs/libstdc++/manual/fstreams.html#std.io.filestreams.binary) on binary files this uses a `filebuf` instead of `istream`, which is purely for formatted input (i.e. textual stuff).
 3. For each group of data in the file (except the header one) we know beforehand the expected size. By checking the size with the one reported by the Fortran format we make the reading process much mode robust.
 4. A new MeshSource is used to simplify the output.

Reviewing order advice:
 - The `SparseUnionFind` first, it is a helper datastructure, but does not relate to the actual content.
 - `FortranUnformattedFile` second, this is the low level reader for the file format
 - The rest. For `centaur.cpp/h`  split-diff is definitely advised, maybe even reading just the new one.

TODO:
- [x] Write some documentation for `FortranUnformattedFile`


applications/Preprocessor/MeshSource.h
```diff
@@ -114,6 +114,74 @@ class MeshSource {
     virtual std::size_t getDimension() const = 0;
 };
 
+/**
```
JensWehner, line 117: soo much better thx

applications/Preprocessor/centaur.cpp
```diff
@@ -257,313 +203,160 @@ CentaurReader::CentaurReader(std::string filename) {
...
-    };
-}
+void CentaurReader::readHeader() {
+    // We do not apriori know the size of the header
+    std::uint32_t headerSize = centaurFile.peekRecordSize();
+    std::vector<char> rawInput(headerSize);
+
+    centaurFile.readRawRecord(headerSize, rawInput.data());
+    std::uint32_t offset = 0;
+    version = *reinterpret_cast<typeof(version)*>(rawInput.data());
```
JensWehner, line 219: I am not a 100% sure about this but I think we should use `memcpy` not `reinterpret_cast` https://www.reddit.com/r/cpp/comments/gb2oz1/convert_bytes_to_object_memcpy_vs_reinterpret/

JensWehner (commented):
Really great changes, I have to look at this again in more detail but it disentangles so much. Nice changes and I learned sth. about disjoint-sets.

applications/Preprocessor/centaur.cpp
Laar: Reading through the documentation on reinterpret_cast here (https://en.cppreference.com/w/cpp/language/reinterpret_cast) it indeed looks that way. As it explicitly notes that you are allowed to cast to `char*` and as well as

> When it is needed to interpret the bytes of an object as a value of a different type, std::memcpy or std::bit_cast (since C++20)can be used

So indeed I have to use memcpy here.

Thanks for catching that.

JensWehner: I recently saw it in a talk, but this is really obscure.

Laar:
Finally found some spare time to finish this. Now good for review and merging.

applications/Preprocessor/lib/centaur.cpp
```diff
@@ -105,52 +108,9 @@ std::string fromFortranString(const std::array<char, N>& fortranString) {
...
-CentaurReader::CentaurReader(std::string filename) {
-    centaurFile.open(filename, std::ios::binary);
-    logger.assert_always(centaurFile.is_open(),
-                         "Cannot open Centaur meshfile.");
-    logger.assert_always(centaurFile.good(),
-                         "Something is not so good about this mesh");
-
+CentaurReader::CentaurReader(std::string filename)
+    : centaurFile(filename), elementCount({0, 0, 0, 0}) {
+    // temp(filename);
```
JensWehner, line 113: remove commented out code

applications/Preprocessor/lib/centaur.cpp
```diff
@@ -257,314 +205,160 @@ CentaurReader::CentaurReader(std::string filename) {
...
-        }
-        logger(VERBOSE, "");
-    };
-    MeshSource::Node node;
-    increment(node);
-    return Range<MeshSource::Node>{node, std::move(increment), numberOfNodes};
+    centaurFile.readRawRecord(headerSize, rawInput.data());
+
+    std::uint32_t offset = 0;
+    memcpy(&version, rawInput.data(), sizeof(offset));
```
JensWehner, line 222: nice

Laar: Yes changed it following your previous comment. Problem with the reinterpret cast was that some of its uses indeed were undefined behaviour, but that the common implementations did do the desired thing.

The commit says that `type*` -> `char*` and then dereferencing is allowed, while `char*` -> `type*` and then dereferencing is not allowed (unless you of course got the `char*` from a `type*`).

JensWehner (commented):
You made this look much nicer. I did not approve yet because I only skimmed through it

applications/Preprocessor/lib/centaur.cpp
Laar: Fixed

JensWehner (approved):
This is a really useful refactor, it makes the whole thing understandable, great work. 

I think a small test where we read-in a centaur file would be helpful, but this is a massive improvement as it is and deserves merging. up to you. :)

JensWehner:
@Laar fell free to merge

PR #101, opened 2021-03-01
Draft: Implement v2 mesh format reading in the kernel

Laar:
A basic implementation of reading the proposed second version of the intermediate mesh format. 

At the moment only reads it in, but discards the data.

Plans:
- [ ] Introduce zones in the mesh structure and apply from the reader
- [ ] Add optional user meta-data to zones (e.g. material parameter)
- [ ] Add tests


JensWehner (approved):
Should it have some tests?

Laar:
> Should it have some tests?

I did test it with one manually updated mesh by running the TutorialPoison problem. But indeed something more structural is needed. I have added that to the TODO list.

However, for you this might already serve to test any output of preprocessor.

Laar:
Made obsolete by #103 and #109.

PR #102, opened 2021-03-02
Preprocessor id safety

Laar:
Introduction of wrappers for the various ids in the preprocessor.

This is a large improvement, but I'm still doubting about it:
 - You could add a dimension template to the EntityGID (i.e. `EntityGId<d>`). Though this requires splitting some methods based on the template argument.
 - I don't really like that I need to repeat almost the same implementation for the three IDs.


JensWehner (commented):
I like this, if you want to reduce boilerplate you can use https://www.fluentcpp.com/2018/04/27/tag-dispatching/

basically make id a template and instantiate 3 templates, with 3 different empty structs ( so that the types are different) and then just use `using CoordId= TemplateId<emptystruct1>`

Laar:
> I like this, if you want to reduce boilerplate you can use https://www.fluentcpp.com/2018/04/27/tag-dispatching/
> 
> basically make id a template and instantiate 3 templates, with 3 different empty structs ( so that the types are different) and then just use `using CoordId= TemplateId<emptystruct1>`

Good suggestion. I applied it, updated it and fixed some errors. It should now be ready to be reviewed & merged.

JensWehner (approved):
Good work!

PR #103, opened 2021-03-04
added new format writer

JensWehner:
Tests are still missing, but I think this works now.


applications/Preprocessor/output_impl.h
```diff
@@ -102,11 +105,20 @@ void outputMesh(Mesh<dimension>& mesh,
     }
     std::ofstream output(outputFileName.getValue());
     output << std::hexfloat;
-    output << "mesh 1" << std::endl;
+    size_t version = 2;
```
Laar, line 108: Can be const

applications/Preprocessor/output_impl.h
```diff
@@ -102,11 +105,20 @@ void outputMesh(Mesh<dimension>& mesh,
...
+    output << "mesh " << version << std::endl;
     output << mesh.getNumberOfNodes() << " " << mesh.getNumberOfElements()
            << " " << dimension;
     printOtherEntityCounts(output, mesh, Detail::tag<dimension - 1>{});
     output << std::endl;
+
+    output << "zones\n";
+    output << mesh.getZoneNames().size() << "\n";
+    for (size_t i = 0; i < mesh.getZoneNames().size(); i++) {
+        output << i << " " << mesh.getZoneNames()[i] << "\n";
```
Laar, line 118: I'm wondering why you output i here? I don't see it in the new spec.

applications/Preprocessor/mesh_impl.h
```diff
@@ -382,7 +382,8 @@ template <std::size_t dimension>
 void Mesh<dimension>::addElement(std::vector<std::size_t> nodeCoordinateIDs,
                                  const std::string& zoneName) {
     std::size_t elementID = elementsList.size();
-    Element<dimension> newElement{this, elementID, 0};
+    size_t zone_id=getZoneId(zoneName);
+    Element<dimension> newElement{this, elementID, zone_id};
```
Laar, line 386: Oops.

Laar (commented):
Looks like a good start.

Thought, I'm wondering why this is going on the `rewrite-centaur-reader` branch. That does not have to do anything with the new intermediate mesh format.

JensWehner:
it has the new meshsource, I maybe was to lazy to find the parent branch

applications/Preprocessor/output_impl.h
JensWehner: ahh I thought the `#` in the spec was the number. Oops

Laar:
> it has the new meshsource, I maybe was to lazy to find the parent branch

But this does not use anything of the new meshsource.

applications/Preprocessor/output_impl.h
```diff
@@ -165,6 +177,8 @@ void outputMesh(Mesh<dimension>& mesh,
         }
         output << std::endl;
     }
+
+    output << "elements\n";
     for (auto element : mesh.getElements()) {
```
Laar, line 182: I don't see the output of which element belongs to which zone here

JensWehner: argh I forgot that part.... oops

Laar (commented):
Looks good to me.

Just a request, could you rebase this so that it goes into the more logical intermediate-mesh-with-zones branch? As the rewrite of the Centaur reader does not have significant priority for me.

Laar:
@JensWehner If you can fix the formatting, then I'll merge it.

Laar:
As this outputs meshes that we can't read yet. For that we need #109, so I think it is best if we merge this directly after that.

applications/Preprocessor/lib/output_impl.h
```diff
@@ -102,11 +103,18 @@ void outputMesh(Mesh<dimension>& mesh,
...
     std::ofstream output(outputFileName.getValue());
     output << std::hexfloat;
-    output << "mesh 1" << std::endl;
+    size_t version = 2;
+    output << "mesh " << version << std::endl;
     output << mesh.getNumberOfNodes() << " " << mesh.getNumberOfElements()
            << " " << dimension;
     printOtherEntityCounts(output, mesh, Detail::tag<dimension - 1>{});
     output << std::endl;
+    output << "zones" << std::endl;
```
Laar, line 112: I think the number of zones should be included here. Something like the following:

```suggestion
    output << "zones" << std::endl;
    output << mesh.getZoneNames().size() << std::endl;
```

Laar (changes requested):
I think I found a difference with the specification.

applications/Preprocessor/lib/output_impl.h
```diff
@@ -72,6 +72,7 @@ void printOtherEntities(std::ofstream& output, const Mesh<dimension>& mesh,
                         tag<d>) {
     for (auto entity : mesh.template getEntities<d>()) {
         std::set<std::size_t> localPartitions;
+        output << "codim" << dimension - d << '\n';
```
Laar, line 75: Codim is now repeated for every edge/face.

```suggestion
    output << "codim" << dimension - d << '\n';
    for (auto entity : mesh.template getEntities<d>()) {
        std::set<std::size_t> localPartitions;
```

Laar (changes requested):
Another problem that I spotted when looking at the actual output. (I must have been sleeping in the initial review).

applications/Preprocessor/lib/output_impl.h
```diff
@@ -102,11 +103,18 @@ void outputMesh(Mesh<dimension>& mesh,
...
            << " " << dimension;
     printOtherEntityCounts(output, mesh, Detail::tag<dimension - 1>{});
     output << std::endl;
+    output << "zones" << std::endl;
+    for (const auto& zonename : mesh.getZoneNames()) {
+        std::cout << zonename << '\n';
+    }
+
+    output << "nodes" << std::endl;
     std::size_t reservedSpace =
```
Laar, line 118: The ordering of the output is incorrect here. It should have been:
 - Partition information (this is old lines 110-119)
 - Zone information

Laar (changes requested):
A problem with the output ordering

Laar (approved):
With some additional fixes it seems to work now.

PR #104, opened 2021-03-04
Expand the correctly calculated owningElements

Laar:
Work from an old branch. Each face, edge and node has an owner. If we own the part then we are certain about the owner, but for parts in the ghost layer we may have insufficient data to determine the owner with certainty. To prevent accidents from reporting the wrong owner, a rather conservative check was added to each of the parts. That is, a fail safe design was used, better report that it is not possible to determine the owner than report the wrong one.

Improved knowledge about the mesh, and ghost layer specifically, allowed widening the number of cases where we could determine the owner of a part with certainty.

### DIFF
--- a/kernel/Base/Face.cpp
+++ b/kernel/Base/Face.cpp
@@ -399,20 +399,16 @@ Element* Face::getOwningElement() const {
         // actually not present.
         std::size_t faceLocalId = elementLeft_->getLocalId(this);
         std::size_t nodes = getReferenceGeometry()->getNumberOfNodes();
-        for (std::size_t i = 0; i < nodes; ++i) {
+        for (std::size_t i = 0; !safe && i < nodes; ++i) {
             std::size_t elementNodeIndex =
                 elementLeft_->getReferenceGeometry()
                     ->getLocalNodeIndexFromFaceAndIndexOnFace(faceLocalId, i);
             const Node* node = elementLeft_->getNode(elementNodeIndex);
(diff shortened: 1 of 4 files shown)


kernel/Base/Face.cpp
```diff
@@ -381,6 +381,51 @@ bool Face::determineIsPeriodicBoundaryFace() const {
...
+        // must be included and we can be certain that the right element is
+        // actually not present.
+        std::size_t faceLocalId = elementLeft_->getLocalId(this);
+        std::size_t nodes = getReferenceGeometry()->getNumberOfNodes();
+        for (std::size_t i = 0; i < nodes; ++i) {
+            std::size_t elementNodeIndex =
+                elementLeft_->getReferenceGeometry()
+                    ->getLocalNodeIndexFromFaceAndIndexOnFace(faceLocalId, i);
+            const Node* node = elementLeft_->getNode(elementNodeIndex);
+            for (const Element* element : node->getElements()) {
```
JensWehner, line 407: could this not be a std::any_of as well?

JensWehner (commented):
Nice extension of the checks, a unit tests would be nice.

Laar:
> Nice extension of the checks, a unit tests would be nice.

Primary hurdle to make any checks for this is that this can only be done on a partitioned mesh. Which requires running it in MPI mode with multiple processes. I don't have an idea how to do that in the current setup.

JensWehner (approved):

Issue #105, opened 2021-03-09
Update Braggstack theory to work in 2D


Laar:
The code that computes the bands of a BraggStack using theoretical means (`BraggStackBandStructure`) only works in 3D. It is desirable to update it to also work in 2D. Doing so would allow better integration/self tests of DGMax, ones where the epsilon is actually varied.

PR #106, opened 2021-03-09
Fix the projection of initial vectors

Laar:
This broke with 08c1ceb43aa1b1f247537c60ddbf567c024783b. Due to not testing different configurations of DGMaxEigenvalue this was not caught earlier.

Note: Some code duplication for printing the tests. I found this acceptable, especially considering that I want to split the test at some point to also use two different structures. (see also #105)


JensWehner (approved):
LGTM

PR #109, opened 2021-04-06
Intermediate mesh with zones kernel

Laar:
Adds reading version 2 of the intermediate format to the kernel, which includes Zone information.

The Zone information is stored in the Mesh. Each element gets a pointer (practically a reference but that has frustrating restrictions on reassigning) to the Zone it belongs to.

Note: The change to the kernel is to fix the structured mesh generator in the kernel. That does not assign zone names, so all the elements would get bogus zone naming.


Laar:
Two thoughts:
 1. I should think of a test for this. Probably have some predefined intermediate mesh in version 2 so that I can check the zone names
 2.  `Zone.cpp` is still empty, this will change when I start using it in DGMax.

kernel/Base/MeshManipulatorBase.cpp
```diff
@@ -90,6 +90,12 @@ MeshManipulatorBase::MeshManipulatorBase(const ConfigurationData* config,
 
 MeshManipulatorBase::MeshManipulatorBase(const MeshManipulatorBase& other) =
     default;
+
+Zone* MeshManipulatorBase::addZone(std::string name) {
+    std::size_t index = zones_.size();
+    zones_.emplace_back(name, index);
+    return &zones_[index];
```
JensWehner, line 97:
```suggestion
    zones_.emplace_back(name, zones_.size());
    return &zones_.back();
```
maybe better, but less clear whar is put into the `emplace_back`

kernel/Base/MeshManipulatorBase.cpp
```diff
@@ -90,6 +90,12 @@ MeshManipulatorBase::MeshManipulatorBase(const ConfigurationData* config,
 
 MeshManipulatorBase::MeshManipulatorBase(const MeshManipulatorBase& other) =
     default;
+
+Zone* MeshManipulatorBase::addZone(std::string name) {
```
JensWehner, line 94: Maybe we can return a reference, because it will not be `nullptr`

kernel/Base/MeshManipulator_Impl.h
```diff
@@ -1119,8 +1119,8 @@ void MeshManipulator<DIM>::addEdgeBasisFunctionSet(
 
 template <std::size_t DIM>
 Base::Element *MeshManipulator<DIM>::addElement(
-    const std::vector<std::size_t> &globalNodeIndexes, std::size_t owner,
-    bool owning) {
+    const std::vector<std::size_t> &globalNodeIndexes, Zone *zone,
```
JensWehner, line 1122: Could we use a reference here or can it be NULL?

kernel/Base/MeshManipulator_Impl.h
```diff
@@ -1536,6 +1555,38 @@ void MeshManipulator<DIM>::readMesh(const std::string &filename) {
         "threads, please rerun the preprocessor first",
         numberOfPartitions, MPIContainer::Instance().getNumberOfProcessors());
 
+    // ZONES //
+    ///////////
+
+    std::vector<Zone *> zones;
+    if (version == 1) {
+        // Zones are only present in version > 1, provide a default zone
+        Zone *zone = addZone("Main");
```
JensWehner, line 1564: This api is a bit counterintuitive for me, we first add a zone and then push it back as well. I would somehow expect the addZone to push_back as well. I am not sure I make sense.

kernel/Base/MeshManipulator_Impl.h
```diff
@@ -1601,7 +1659,7 @@ void MeshManipulator<DIM>::readMesh(const std::string &filename) {
         input >> partition;
         Base::Element *element = nullptr;
         if (partition == processorID) {
-            element = addElement(coordinateIndices, partition, true);
+            element = addElement(coordinateIndices, nullptr, partition, true);
```
JensWehner, line 1662: I would give the nullptr a name,
```suggestion
*zone default=nullptr;
bool owning=true;
            element = addElement(coordinateIndices, default, partition, owning);
```

JensWehner (approved):
Small API suggestions. Otherwise nice change. As you mentioned some tests would be good. GOod work.

kernel/Base/MeshManipulator_Impl.h
Laar: It will (unfortunately) be null in the reader. I would like it differently, but that would require quite some restructuring.

kernel/Base/MeshManipulator_Impl.h
Laar: It is just to have a simple list of pointers, so that I can later easily do `zones[index]` instead of `&(getZones()[index])` or something similar. I'll add a comment to it to clarify it.

kernel/Base/MeshManipulatorBase.cpp
Laar: Done

kernel/Base/MeshManipulatorBase.cpp
Laar: The `&zones_.back()` is quite nice, so I'll use that

Laar (commented):
Implemented part of the feedback

PR #110, opened 2021-04-13
Fix zone assignment in preprocessor

Laar:
Fixes to problems found when trying to use #103 and #109 together to use the new mesh format.

Fixes:
 - The zone was passed from the MeshSource to addElement(), but the latter did not actually use it.
 - The elements generated by the structured mesh generator did not have any zone information.

As backup to the second problem a check is placed that checks if an actual zone is present.


JensWehner (approved):
Good catch

PR #111, opened 2021-04-15
Add more mesh statistics in Preprocessor

Laar:
These should more easily allow observing if a mesh is properly periodic. Additionally a listing of zones can be useful for debugging zone assignments.


JensWehner (approved):
That is simply a very good addition.

PR #113, opened 2021-04-22
Workaround for petsc bug

Laar:
Workaround for a [bug in PETSc](https://gitlab.com/petsc/petsc/-/merge_requests/3901). Instead of inserting the matrix block as a whole, insert each entry and suppress the zero entries that cause the problem.

Note the `==0.0` is intentional, only exact zeros should be (but are not) ignored by PETSc and thus cause the problem.


JensWehner (approved):

JensWehner:
What an interesting bug.

PR #114, opened 2021-05-11
Preprocessor abstractions and documentation

Laar:
The preprocessor has it's own mesh and element shape classes. There are two issues that this PR improves in these classes:

 1. The element shape class was practically undocumented
 2. How to store references to lower dimensional entities when these entities have types indexed by their dimension.
 3. Reducing complexity of ElementShape
 4. Seperate ElementShape definition from listing of shapes

The problem in point 2 stems from that the `ElementShape` (e.g. the hpgem triangle) and `MeshEntity` (e.g. an edge in the mesh) types have their own dimension as template argument. A mesh storing all the MeshEntities would thus need to store `MeshEntity`s of dimension 0, 1, up to and including the dimension of the mesh. Similarly if an `ElementShape` needs to store the shapes that comprise its boundary. This can't be done in a regular array, as the entities and shapes of different dimensions have different types due to the dimension being a template parameter. A custom way to store these was implemented for both in the form of `MeshEntities` and `ElementEntities` structs. In this PR we introduce a new class `TemplateArray` that abstracts the pattern of storing an array of the form `[V<0>, V<1>, ... V<size-1>]`  where `V<size_t>` is a template with an integer index.

Point 3 has several small but substantial improvements:
1. Remove the use of codimensions as this requires specializing the function differently for negative values from positive ones
2. Use -1 as base case for several dimension recursion functions. As stopping at 0 means that one needs to duplicate the code for the d-dimensional case to the 0-dimensional case.
3. Introduced an intermediate class `ElementShapePart`, to reduce the large numbers of unnamed vector nesting. This results in that for the connectivity data the indices for the entity index and target dimension have been switched.

Point 4 is the reason for the new ElementShapes files. While I still don't like the global variable, I think it is an improvement over the old one, and that it nicely splits the concept of a shape from the available shapes.


applications/Preprocessor/lib/TemplateArray.h
```diff
@@ -0,0 +1,150 @@
...
+ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef HPGEM_TEMPLATEARRAY_H
+#define HPGEM_TEMPLATEARRAY_H
+
+#include "tag.h"
+
+namespace Preprocessor {
+
+template <std::size_t size, template <std::size_t> class V>
```
JensWehner, line 45: I see what you are doing here, but could we use std::tuple instead?

applications/Preprocessor/lib/elementShape.h
```diff
@@ -52,124 +57,118 @@ class ElementShape;
...
 
     template <typename... Args>
-    ElementShape(Args... args) : shapeData(args...) {
-        completeSubShapes(tag<dimension - 1>{}, tag<dimension - 1>{});
+    ElementShape(Args... args) : shapeParts({}, args...) {
+        // Insert ElementShapePart corresponding to the whole element.
+        auto& shapeVec = shapeParts.template get<dimension>();
+        std::vector<std::size_t> vertices(getNumberOfNodes());
+        std::iota(vertices.begin(), vertices.end(), 0);
+        shapeVec.emplace_back(
```
JensWehner, line 168: can be a push_back I think

applications/Preprocessor/lib/tag.h
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * Which can't be specialized, as the template argument is not present in the
+ * argument list, nor return type. To allow specialization add the tag:
+ *
+ * template<std::size_t dimension>
+ * void function(non-template-arguments, tag<dimension>)
+ *
+ * @tparam i The integer to pass.
+ */
+template <std::size_t i>
```
JensWehner, line 61: this is nice

JensWehner (commented):
I wait with full review when the full PR is here. I like the simplifications in general. 

To be honest I think more and more the templating of the dimension was not a good idea in general it makes the code complicated for little gain, but it is as it is. 

Otherwise nice work.

applications/Preprocessor/lib/TemplateArray.h
Laar: If the size was constant I think I could, but the use case the size is the dimension (+1). So then I would need to vary the size of the tuple depending on the dimension, which I would not know how. Moreover, I would somehow need to supply size-1, size-2, etc. as template argument. 

So the basic answer is, I would not know of a good and cleaner way. At least not one that is cleaner than this approach.

applications/Preprocessor/lib/elementShape.h
Laar: Yes, or I should remove the constructor.

Laar: Updated

Laar:
As good as I want to make it now considering the amount of time involved and the possible gain there is.


> To be honest I think more and more the templating of the dimension was not a good idea in general it makes the code complicated for little gain, but it is as it is.

I'm more mixed on this, problem is more that c++ does not really support it that nicely. The more important 'bad' idea was in my opinion the two seperate definitions of the reference shapes, one in the kernel and a stripped/added version in the Preprocessor. Removing the latter is quite a bit more complicated due to the dimension-templating and by having more topological data than the kernel. So while I would like to unify them, I don't think that can be done without quite some time investment. Investment without little gain at the moment.

applications/Preprocessor/lib/ElementShapes.h
```diff
@@ -0,0 +1,57 @@
...
+#include "TemplateArray.h"
+
+namespace Preprocessor {
+
+namespace Detail {
+// Alias to have a template type that directly depends on the dimension
+template <std::size_t dimension>
+using ShapePointerVec = std::vector<const ElementShape<dimension>*>;
+}  // namespace Detail
+
```
JensWehner, line 51: what is the purpose of this?

applications/Preprocessor/lib/ElementShapes.cpp
```diff
@@ -0,0 +1,183 @@
...
+const ElementShape<2> square{
+    // Boundary lines
+    Detail::generateEdges<2>({{0, 1}, {0, 2}, {1, 3}, {2, 3}}),
+    // Corner vertices
+    Detail::generateVertices<2>(4)};
+
+const ElementShape<3> tetrahedron{
+    // Faces
+    std::vector<Detail::ElementShapePart<2, 3>>{
+        Detail::ElementShapePart<2, 3>{&triangle, {0, 3, 2}},
```
JensWehner, line 81: does the order matter, I guess yes? Can we add a reference to what these indices point to?

applications/Preprocessor/lib/ElementShapes.h
Laar: The next line needs it `TemplateArray<4, Detail::ShapePointerVec> hpgemShapes`. The `TemplateArray` expects as second template argument a template type 'T<std::size_t>'. Doing `TemplateArray<4,std::vector<const ElementShape<TEMPLATE ARGUMENT>*>` does not  work as the template argument is needed somewhere in the middle. By generating an alias you 'move' the argument so that you have a template name that directly depends on the argument.

applications/Preprocessor/lib/TemplateArray.h
```diff
@@ -0,0 +1,150 @@
...
+    TemplateArray() = default;
+    ~TemplateArray() = default;
+    TemplateArray(const TemplateArray&) = default;
+    TemplateArray(TemplateArray&&) noexcept = default;
+    TemplateArray& operator=(const TemplateArray&) = default;
+    TemplateArray& operator=(TemplateArray&&) noexcept = default;
+};
+
+template <std::size_t>
+class Sample;
```
JensWehner, line 146: What is this for?

applications/Preprocessor/lib/mesh.h
```diff
@@ -619,8 +557,8 @@ class Mesh {
     /// pairs of (adjecent Element, localIndex) that the MeshEntity is indeed at
     /// localIndex of the Element.
     template <std::size_t d>
-    bool checkEntities(tag<d>) const {
-        for (auto entity : otherEntities.template getData<d>()) {
+    bool checkEntities(tag<d> dimTag) const {
+        for (auto entity : otherEntities[dimTag]) {
             for (std::size_t i = 0; i < entity.getNumberOfElements(); ++i) {
```
JensWehner, line 562: would be a place for std::any_of but outside the scope of the PR

applications/Preprocessor/lib/mesh.h
```diff
@@ -677,8 +615,8 @@ class Mesh {
     template <std::size_t entityDimension>
     std::size_t newEntity() {
         std::size_t newIndex =
-            otherEntities.template getData<entityDimension>().size();
-        otherEntities.template getData<entityDimension>().push_back(
+            otherEntities.template get<entityDimension>().size();
+        otherEntities.template get<entityDimension>().push_back(
```
JensWehner, line 619: nicer

applications/Preprocessor/lib/ElementShapes.cpp
Laar: Documented the constructor a bit further to clarify the information. The ordering matters, as it defines which triangle node matches with which node of the tetrahedron.

applications/Preprocessor/lib/mesh.h
```diff
@@ -689,10 +627,14 @@ class Mesh {
 
     std::vector<Element<dimension>> elementsList;
     std::vector<coordinateData> coordinates;
+
+    // Alias template to fill in the dimension of the mesh
+    template <std::size_t entityDimension>
+    using MeshEntityVec = std::vector<MeshEntity<entityDimension, dimension>>;
     // re-stores slices of the elements for consistent use of getEntities<d>,
```
JensWehner, line 634: comment is out of data now ?

applications/Preprocessor/lib/elementShape_impl.h
```diff
@@ -163,34 +74,43 @@ std::enable_if_t<(entityDimension >= 0 && entityDimension < dimension &&
...
+    return shapeParts.template get<entityDimension>()[entityIndex]
+        .getAdjacentShapes(targetDimension);
 }
 
 template <std::size_t dimension>
-template <std::size_t d, std::size_t shapeDimension>
+template <int d, std::size_t shapeDimension>
 bool ElementShape<dimension>::checkSingleShape(
     const ElementShape<shapeDimension>* boundingShape,
     std::vector<std::size_t> boundaryNodes, std::size_t currentIndex,
```
JensWehner, line 85: we need the copy here?

JensWehner: out of scope of the PR

applications/Preprocessor/lib/mesh.h
Laar: No, but it is now improved.

applications/Preprocessor/lib/elementShape_impl.h
Laar: What copy do you mean?

applications/Preprocessor/lib/ElementShapes.cpp
JensWehner: Ahh so it refers to the hpgem ordering.

JensWehner: I think that should be mentioned somewhere with a link to the ascii pictures

JensWehner (changes requested):
The architecture decisions make a lot of sense to me and the code is cleaner. Adding a unittest to TemplateArray is for me the only thing that is missing apart from small bits of documentation Great work.

applications/Preprocessor/lib/elementShape_impl.h
JensWehner: boundaryNotes. Sorry.

applications/Preprocessor/lib/ElementShapes.cpp
Laar: Done

applications/Preprocessor/lib/TemplateArray.h
Laar: As test (and now it's gone)

applications/Preprocessor/lib/elementShape_impl.h
Laar: Does not seem like it, but as outside and it is only called like a few times (no time usage) I don't it is worth fixing it now.

JensWehner (approved):
Looking good, great work!

PR #115, opened 2021-05-12
Update all test meshes to the V2 format

Laar:
Update the test meshes using the `regenerateMeshFiles` script to get them to version 2 of the intermediate mesh format.

This allows for much easier testing of any changes in the preprocessor, as one can directly observe the differences from a modified preprocessor.


JensWehner (approved):
Good idea thanks for doing that.

The new mesh format is soo much easier to read

PR #118, opened 2021-06-18
Another missing import, but now in a less used part of the code

Laar:


JensWehner (approved):

JensWehner:
any idea why the CI did not pick it up?

Laar:
> any idea why the CI did not pick it up?

I expect because of the version of the compiler/standard library, where the newer version probably has probably cleaner headers.

I have here GCC 11.1, while Ubuntu uses 10 (for 21.04) and 9 (20.04 LTS).

PR #119, opened 2021-06-21
Update Basic.yml

JensWehner:
So far our CI only tested gcc and not clang


Laar:
Looks good to me

Laar:
@JensWehner I have updated the required checks as well. Let's hope I did it correctly

PR #121, opened 2021-06-22
Abstract dimensionless base class

Laar:
Introduce a class that abstracts over the common pattern:
```
template<std::size_t DIM>
class T;

class TBase {
  T<0>& operator();
  T<1>& operator();
  // etc.
  // Possible extra methods
};

template<std::size_t DIM>
class T : public TBase{
 // ...
};
```

Currently this is used for the two different Point classes. With this added base class the pattern is more formalized and allows for easier uses at different places.


JensWehner (approved):
So this is basically CRTP.

PR #122, opened 2021-06-23
Fixed dimension ReferenceToPhysicalMapping (uses #121)

Laar:
This changes the mapping between Reference and Physical space. That is, the transformations used to convert between the elements in the mesh (triangles, tetrahedra, etc.) and their reference counter parts which are used to define the basis functions.

This PR replaces the old interface that was a specialization of `MappingInterface<0>` that had methods for dimensions 1-4. It is replaced by a version that is templated on the dimension argument plus a base class without the dimension argument. This is in my view a much cleaner and flexible way to define these mappings, as it does only have the methods for the right dimension.

Reviewing advice:
 1. Look at #121, this code uses the new class introduced there
 2. Look at `MappingReferenceToPhysical`
 3. The rest is refactoring to match the new mapping interface.

### DIFF
--- a/kernel/Geometry/Mappings/MappingToPhysSimplexLinear.h
+++ b/kernel/Geometry/Mappings/MappingToPhysSimplexLinear.h
@@ -69,14 +69,17 @@ class MappingToPhysSimplexLinear : public MappingReferenceToPhysicalDim<DIM> {
         reinit();
     }
 
-    MappingToPhysSimplexLinear(const MappingToPhysSimplexLinear<DIM>& other)
-        : MappingReferenceToPhysical(other) {}
+    MappingToPhysSimplexLinear(const MappingToPhysSimplexLinear<DIM>& other) =
+        default;
 
     PointPhysical<DIM> transform(const PointReference<DIM>&) const final;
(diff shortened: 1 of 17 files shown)


kernel/Geometry/Mappings/MappingToPhysSimplexLinear.h
```diff
@@ -62,22 +62,24 @@ namespace Geometry {
...
-        : MappingReferenceToPhysical(pG) {
-        logger.assert_debug(pG != nullptr, "Invalid physical geometry passed");
+        : MappingReferenceToPhysical<DIM>(pG) {
         reinit();
     }
 
-    MappingToPhysSimplexLinear(const MappingToPhysSimplexLinear<DIM>& other)
-        : MappingReferenceToPhysical(other) {}
+    MappingToPhysSimplexLinear(const MappingToPhysSimplexLinear<DIM>& other) =
+        default;
```
JensWehner, line 73: nice

JensWehner (approved):
Pretty clever, I like the whole thing just for the simplification alone. 

@Laar good job.

PR #123, opened 2021-06-29
Remove negative quadratures

Laar:
Removes several negative weight quadrature rules as these are less numerically stable. This stability is especially important when computing error integrals `||(u_theory - u_computed)^2||`. Such an integral should always be positive, but with negative weights the rounding might make the integrand negative. 

Therefore this PR removes the negative weight rules. Some replacements are added for tetrahedra. Previous quadrature rules were all individually derived from the quadrature base class making them tedious to implement. To simplify the implementation of quadrature rules the  `StandardGaussQuadratureRule` is added.

Note automatic formatting is turned off to improve the readability of the tables with quadrature weights & points.

Reviewing order advice: Look at `StandardQuadratureRule` before looking at the changes to the Tetrahedron quadrature rules.


JensWehner (approved):
So far looking good, I am not a big friend of singeltons but for this it should be more than fine.

JensWehner (commented):
Architecture looks fine to me.

Laar:
@JensWehner Finally ready (had to fix a cache memory leak in QuadratureRules). So a final look would be appreciated.

tests/self/TestMeshes.h
```diff
@@ -160,17 +160,20 @@ std::vector<std::string> getUnitCubeCubeMeshes(
...
+    logger.assert_always(
+        minLevel >= 2 && maxLevel >= 2,
+        "Periodic meshes with 1 or 2 elements are not supported");
     std::string prefix = getCMAKE_hpGEM_SOURCE_DIR() + "/tests/files/";
     return limit(
         {
-            prefix + "unitCubeN1Per.hpgem",
-            prefix + "unitCubeN2Per.hpgem",
+            //            prefix + "unitCubeN1Per.hpgem",
+            //            prefix + "unitCubeN2Per.hpgem",
```
JensWehner, line 171: remove commented out code

JensWehner (approved):
Looking good, small comment removal otherwise very nice.

tests/self/TestMeshes.h
Laar: Done so by motivating why these meshes are absent. By leaving them commented out it is clear why/that we only start at level 2.

Laar:
Currently CI fails due to the issue fixed in #131. As it all ran correctly and I reformatted the code locally (no changes) I will merge this manually.

PR #125, opened 2021-07-03
Add zone based material information into DGmax

Laar:
Improve and replace the way the material coefficients (i.e. epsilon) are assigned in DGMax.

The way these values are now computed using the strategy pattern in the form of `StructureDescription`. There are two implementations:
 1. The old structures in `PredefinedStructure`, both for legacy reasons and to allow testing with structured meshes that lack zone information 
 2. Based on the zones in the mesh file using `ZoneStructureDescription`, which uses regexes to match the name of the zone to the material parameters.

The DGMaxProgramUtils has a method that converts a simple string input heuristically into the right implementation.

TODO before merging:
 - [x] Clean up of excess commnets
 - [x] Look into testing some of the code (e.g. the ZoneStructureDescription).

Recommended reviewing order:
 1. First `StructureDescription` to understand the new class
 2. The two implementations `PredefinedStructure` and `ZoneStructureDescription`
 3. The rest


applications/DG-Max/DGMaxProgramUtils.cpp
```diff
@@ -148,4 +149,76 @@ PointPath<DIM> parsePath(const std::string& path) {
...
+                                                         zoneEpsilons);
+}
+
+std::unique_ptr<StructureDescription> determineStructureDescription(
+    const std::string& input, std::size_t dim) {
+
+    // Test to see if the input is a number, by trying to parse it
+    char* end;
+    long value = std::strtol(input.c_str(), &end, 10);
+    if (!*end) {
```
JensWehner, line 198: Can we make this a small function "convertToLong" or so

applications/DG-Max/DGMaxProgramUtils.h
```diff
@@ -49,6 +49,22 @@ struct PointPath {
...
+/// Heuristically determine the way to determine the structure description.
+/// It accepts three forms:
+///   - A single integer: the index into the predefined structures
+///   - A filename: The file should contain lines of the form 'regex,epsilon',
+///     where regex is a regular expression that matches some of the zones and
+///     epsilon is the corresponding epsilon.
+///   - An inline version of the filename using where each regex,epsilon pair is
+///     separated by a semicolon. E.g. regex1,1;regex2,12.1
+///
+/// \param input The index or file name
```
JensWehner, line 62: I do not understand the API completely, it takes a string as input. So the string can be a nuber a filename, What is an inline filename? I think I do not fully understand the problem yet

applications/DG-Max/Utils/PredefinedStructure.h
```diff
@@ -0,0 +1,78 @@
...
+        : structure_(structure), dimension(dimension){};
+
+    ElementInfos* createElementInfo(const Base::Element* element) final;
+
+   private:
+    template <std::size_t DIM>
+    ElementInfos* createElementInfoDim(const Base::Element* element) const;
+
+    PredefinedStructure structure_;
+    std::size_t dimension;
```
JensWehner, line 74:
```suggestion
    std::size_t dimension_;
```

applications/DG-Max/Utils/PredefinedStructure.h
```diff
@@ -0,0 +1,78 @@
...
+PredefinedStructure structureFromInt(std::size_t value);
+
+/// Some predefined structures.
+///
+/// Both for historical meshes without zone information and structured meshes
+class PredefinedStructureDescription : public StructureDescription {
+   public:
+    PredefinedStructureDescription(PredefinedStructure structure,
+                                   std::size_t dimension)
+        : structure_(structure), dimension(dimension){};
```
JensWehner, line 65:
```suggestion
        : structure_(structure), dimension_(dimension){};
```

applications/DG-Max/Utils/ZoneStructureDescription.cpp
```diff
@@ -0,0 +1,53 @@
...
+ */
+#include "ZoneStructureDescription.h"
+namespace DGMax {
+
+ElementInfos *ZoneInfoStructureDefinition::createElementInfo(
+    const Base::Element *element) {
+    auto &zone = element->getZone();
+    int index = zone.matchName(regexes_);
+    if (index >= 0) {
+        return new ElementInfos(epsilons_[index]);
```
JensWehner, line 46: Can we not somehow make this into `unique_ptr`? Probably a large scale refactor.

applications/DG-Max/Utils/ZoneStructureDescription.h
```diff
@@ -0,0 +1,85 @@
...
+    ///
+    /// \param regexes vector of regexes used to match zone names
+    /// \param epsilons vector with corresponding values of epsilon
+    /// \param defaultEpsilon Optional default epsilon when no regex matches
+    ///   (nan, default, will generate an error)
+    ZoneInfoStructureDefinition(std::vector<std::regex> regexes,
+                                std::vector<double> epsilons,
+                                double defaultEpsilon = std::nan(""))
+        : regexes_(std::move(regexes)),
+          epsilons_(std::move(epsilons)),
```
JensWehner, line 69:
```suggestion
        : regexes_(regexes),
          epsilons_(epsilons),
```

I would argue it does not matter here and the compiler will optimise it away anyway.

applications/DG-Max/Utils/PredefinedStructure.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+ElementInfos* PredefinedStructureDescription::createElementInfo(
+    const Base::Element* element) {
+    switch (dimension) {
+        case 2:
+            return createElementInfoDim<2>(element);
+        case 3:
+            return createElementInfoDim<3>(element);
+        default:
+            logger.assert_always(false, "Not programmed for dimension %",
```
JensWehner, line 51:
```suggestion
            logger.assert_always(false, "Not implemented for dimension %",
```

applications/DG-Max/Programs/DGMaxEigenConvergence.cpp
```diff
@@ -69,9 +69,9 @@ auto &method = Base::register_argument<std::string>(
 /// Create a reference bandstructure based on the structure index
 template <std::size_t DIM>
 std::unique_ptr<BandStructure<DIM>> createStructure(
-    std::size_t structureIndex) {
+    DGMax::PredefinedStructure structureIndex) {
     // Using same structure indices as for the Jelmer structure
-    if (structureIndex == 0) {
+    if (structureIndex == DGMax::PredefinedStructure::VACUUM) {
```
JensWehner, line 74: nice!

applications/DG-Max/DGMaxProgramUtils.cpp
```diff
@@ -148,4 +149,76 @@ PointPath<DIM> parsePath(const std::string& path) {
 template PointPath<1> parsePath(const std::string& path);
 template PointPath<2> parsePath(const std::string& path);
 template PointPath<3> parsePath(const std::string& path);
+
+std::unique_ptr<ZoneInfoStructureDefinition> readZonedDescription(
```
JensWehner, line 153: Could you add a small example of such a file as a comment, please? :)

JensWehner (commented):
Looking good, so far

applications/DG-Max/DGMaxProgramUtils.cpp
Laar: What should it do beyond what `std::strtol` is already doing?

Note to self: I should reread the docs on it and fix the end pointer.

applications/DG-Max/Utils/ZoneStructureDescription.cpp
Laar: Might indeed be a quite significant refactor, as a pointer to the base type (`UserElementData`) is stored in at least every element.

applications/DG-Max/DGMaxProgramUtils.cpp
JensWehner: Yes, but it is not really nice to read because of all the pointers and basically you want to convert a string to a long

applications/DG-Max/Utils/ZoneStructureDescription.h
Laar: From what I have been reading online it is not included in the valid locations where it is allowed to optimize this away.

Apart from the technical side, this is just what clang-tidy suggests for it. So I assume there is a good reason to do this (or at least no bad side in not doing it).

applications/DG-Max/DGMaxProgramUtils.h
Laar: Should be clearer now, but please check.

applications/DG-Max/Utils/PredefinedStructure.h
Laar: Done via refactor.

applications/DG-Max/DGMaxProgramUtils.cpp
Laar: Now included in the documentation of `determineStructureDescription`

applications/DG-Max/DGMaxProgramUtils.h
JensWehner: Yes it is thanks

JensWehner (approved):
Nice refactor well done. This code is becoming more and more readable thanks to your efforts, well done.

Laar:
> Nice refactor well done. This code is becoming more and more readable thanks to your efforts, well done.

Thank you. I just added a simple test for the `ZoneStructureDescription` and remove some excessive commented code. So this should be ready to merge if the tests succeed.

PR #127, opened 2021-07-05
Update gmsh.cpp

JensWehner:
Should fix the failing PBC reading in, it is a bit of a crude fix, but it should be good enough atm.

### DIFF
--- a/applications/Preprocessor/lib/gmsh.cpp
+++ b/applications/Preprocessor/lib/gmsh.cpp
@@ -387,6 +387,8 @@ void GmshReader::readPBCs() {
 
     bool found_PBCs = locate_in_file(Filehandle_, "$Periodic");
     if (!found_PBCs) {
+        Filehandle_.clear();
+        Filehandle_.seekg(0);
         return;
     }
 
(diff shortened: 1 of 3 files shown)


Laar (approved):
Simple and effective!

JensWehner:
@Laar I reread the documentation again and what I got from it is that the order of sections is fixed. You mentioned that sections can appear multiple times, do you have source? I probably overlooked it.

Laar:
@JensWehner 
It is stated for the new format, but I don't see it for the legacy format in the current documentation. Specifically for the newer format they state:
> Sections can be repeated in the same file, and post-processing sections can be put into separate files

Using the  [wayback machine](https://web.archive.org/web/20160408183126/http://gmsh.info/doc/texinfo/gmsh.html#MSH-ASCII-file-format) I can find the original documentation


>When $Elements are given, $Nodes should also be provided, before the $Elements section. Currently only one $Nodes and one $Elements section are allowed per file. (This might/will change in the future.)
>
> [ommitted]
>
> Sections can be repeated in the same file, and post-processing sections can be put into separate files (e.g. one file per time step). Nodes are assumed to be defined before elements.

PR #129, opened 2021-07-14
Introduce an abstraction over MeshEntities

Laar:
For several processes in hpgem we need to handle all parts of the Mesh (Element, Face, Edge and Node) in an (almost) identical fashion. The current implementations do this by repeating the code for each part, which is both error prone and ugly. This PR tries to alleviate these code smells.

Specifically this PR:
 1. Introduces a pure virtual base class `MeshEntity`, that is the parent of each if the four parts of the Mesh
 2. Introduces a simple visitor to allow differentiating behavior between the different MeshEntities.
 3. Rewrites several parts of `GlobalIndexing` as a proof of concept that it works and the possible effect on code quality.

Some attempts have been made to ensure that the compiler can (theoretically) devirtualize calls to MeshEntities. This is done by:
 1. Marking all the implementation of the functions of `MeshEntity` in the parts as final
 2. Exposing methods that accept a MeshEntity to allow inlining, thereby allowing further differentiation based on the actual MeshEntity type.

First review request: What do you think about the idea/implementation? Any suggestions on whether to continue with this idea?

Natural reviewing order is to start with `MeshEntity` and proceed to either the parts themselves (rather straight forward) or the usage in `GlobalIndexing`.

### DIFF
--- a/kernel/Base/Edge.cpp
+++ b/kernel/Base/Edge.cpp
@@ -111,7 +111,7 @@ bool Edge::isOwnedByCurrentProcessor() const {
     return elements_.size() > 0 && elements_[0]->isOwnedByCurrentProcessor();
 }
 
-Element* Edge::getOwningElement() const {
+const Element* Edge::getOwningElement() const {
 #if HPGEM_ASSERTS
     if (!isOwnedByCurrentProcessor()) {
         // This edge is part of the layer of ghost cells around the owned part
--- a/kernel/Base/Edge.h
+++ b/kernel/Base/Edge.h
@@ -42,6 +42,7 @@
 #include <vector>
 #include <cstdlib>
 
+#include "MeshEntity.h"
 #include "Logger.h"
 #include "TreeEntry.h"
 
@@ -58,7 +59,7 @@ class Element;
  * so they can connect edge-based conforming degrees of freedom to the proper
  * elements. \todo 4D support
  */
-class Edge {
+class Edge : public MeshEntity {
    public:
     explicit Edge(std::size_t ID)
         : numberOfConformingDOFOnTheEdge_(std::vector<std::size_t>(0, 0)),
@@ -174,6 +175,14 @@ class Edge {
     /// The element owning this edge, only valid if the edge is owned by the
     /// current processor
     Element* getOwningElement() const;
+
+    void visitEntity(MeshEntityVisitor<>& visitor) final {
+        visitor.visit(*this);
+    }
+
+    void visitEntity(MeshEntityVisitor<true>& visitor) const final {
+        visitor.visit(*this);
+    }
 
    private:
     std::vector<Element*> elements_;
--- a/kernel/Base/Edge.h
+++ b/kernel/Base/Edge.h
@@ -71,7 +71,7 @@ class Edge : public MeshEntity {
 
     void addElement(Element* element, std::size_t edgeNumber);
 
-    std::size_t getLocalNumberOfBasisFunctions() const {
+    std::size_t getLocalNumberOfBasisFunctions() const final {
         std::size_t number = numberOfConformingDOFOnTheEdge_[0];
         for (std::size_t index : numberOfConformingDOFOnTheEdge_)
             logger.assert_debug(
@@ -80,7 +80,8 @@ class Edge : public MeshEntity {
         return numberOfConformingDOFOnTheEdge_[0];
     }
 
-    std::size_t getLocalNumberOfBasisFunctions(std::size_t unknown) const {
+    std::size_t getLocalNumberOfBasisFunctions(
+        std::size_t unknown) const final {
         // TODO: LC, numberOfConformingDOFOnTheNode_ might be smaller than
         // the number of unknowns (as that is not known here). Thus we might
         // index beyond the number of unknowns.
@@ -88,7 +89,7 @@ class Edge : public MeshEntity {
         return numberOfConformingDOFOnTheEdge_[unknown];
     }
 
-    std::size_t getTotalLocalNumberOfBasisFunctions() const {
+    std::size_t getTotalLocalNumberOfBasisFunctions() const final {
         std::size_t result = 0;
         for (auto nbasis : numberOfConformingDOFOnTheEdge_) result += nbasis;
         return result;
@@ -100,7 +101,7 @@ class Edge : public MeshEntity {
         return getLocalNumberOfBasisFunctions();
     }
 
-    std::size_t getID() const { return ID_; }
+    std::size_t getID() const final { return ID_; }
 
     ///\deprecated Does not conform naming conventions, use getNumberOfElements
     /// instead
@@ -170,11 +171,11 @@ class Edge : public MeshEntity {
         return positionInTheTree_;
     }
 
-    bool isOwnedByCurrentProcessor() const;
+    bool isOwnedByCurrentProcessor() const final;
 
     /// The element owning this edge, only valid if the edge is owned by the
     /// current processor
-    Element* getOwningElement() const;
+    const Element* getOwningElement() const final;
 
     void visitEntity(MeshEntityVisitor<>& visitor) final {
         visitor.visit(*this);
--- a/kernel/Base/Element.h
+++ b/kernel/Base/Element.h
@@ -56,6 +56,8 @@
 #include <memory>
 #include <Integration/QuadratureRules/GaussQuadratureRule.h>
 #include <Base/ElementBasisFunctions.h>
+#include "MeshEntity.h"
+
 namespace hpgem {
 
 namespace FE {
@@ -74,7 +76,9 @@ class PhysicalElement;
 
 // class is final as a reminder that the virtual default destructor should be
 // added once something inherits from this class
-class Element final : public Geometry::ElementGeometry, public ElementData {
+class Element final : public Geometry::ElementGeometry,
+                      public ElementData,
+                      public MeshEntity {
    public:
     using SolutionVector = LinearAlgebra::MiddleSizeVector;
     using CollectionOfBasisFunctionSets =
@@ -527,6 +531,14 @@ class Element final : public Geometry::ElementGeometry, public ElementData {
     void setOwnedByCurrentProcessor(std::size_t owner, bool owned);
     bool isOwnedByCurrentProcessor() const;
     std::size_t getOwner() const;
+
+    void visitEntity(MeshEntityVisitor<>& visitor) final {
+        visitor.visit(*this);
+    }
+
+    void visitEntity(MeshEntityVisitor<true>& visitor) const final {
+        visitor.visit(*this);
+    }
 
     /// Output operator.
     friend std::ostream& operator<<(std::ostream& os, const Element& element);
--- a/kernel/Base/Element.h
+++ b/kernel/Base/Element.h
@@ -98,9 +98,7 @@ class Element final : public Geometry::ElementGeometry,
 
     Element* copyWithoutFacesEdgesNodes();
 
-    std::size_t getID() const;
-
-    std::size_t getID();
+    std::size_t getID() const final { return id_; }
 
     void setQuadratureRulesWithOrder(std::size_t quadrROrder);
 
@@ -380,15 +378,16 @@ class Element final : public Geometry::ElementGeometry,
     /// element only. This always includes functions with compact support on the
     /// interior of the element and DG basis function, but never include
     /// conforming basis functions that are nonzero on a face, edge or node
-    std::size_t getLocalNumberOfBasisFunctions() const {
+    std::size_t getLocalNumberOfBasisFunctions() const final {
         return basisFunctions_.getNumberOfLocalBasisFunctions();
     }
 
-    std::size_t getLocalNumberOfBasisFunctions(std::size_t unknown) const {
+    std::size_t getLocalNumberOfBasisFunctions(
+        std::size_t unknown) const final {
         return basisFunctions_.getNumberOfLocalBasisFunctions(unknown);
     }
 
-    std::size_t getTotalLocalNumberOfBasisFunctions() const {
+    std::size_t getTotalLocalNumberOfBasisFunctions() const final {
         return basisFunctions_.getTotalLocalNumberOfBasisFunctions();
     }
 
@@ -529,8 +528,10 @@ class Element final : public Geometry::ElementGeometry,
     /// \param owned Whether this element is owned or not by the current
     /// processor.
     void setOwnedByCurrentProcessor(std::size_t owner, bool owned);
-    bool isOwnedByCurrentProcessor() const;
+    bool isOwnedByCurrentProcessor() const final;
     std::size_t getOwner() const;
+
+    const Element* getOwningElement() const final { return this; }
 
     void visitEntity(MeshEntityVisitor<>& visitor) final {
         visitor.visit(*this);
--- a/kernel/Base/Face.h
+++ b/kernel/Base/Face.h
@@ -44,6 +44,7 @@
 #include "Base/FaceMatrix.h"
 #include "Geometry/FaceGeometry.h"
 #include "Base/Element.h"
+#include "MeshEntity.h"
 #include "L2Norm.h"
 #include "TreeEntry.h"
 
@@ -61,7 +62,9 @@ struct FaceCacheData;
 /// FaceGeometry holds all FaceReference related data and appropriate mappings
 // class is final as a reminder that the virtual default destructor should be
 // added once something inherits from this class
-class Face final : public Geometry::FaceGeometry, public FaceData {
+class Face final : public Geometry::FaceGeometry,
+                   public FaceData,
+                   public MeshEntity {
    public:
     using FaceQuadratureRule = QuadratureRules::GaussQuadratureRule;
 
@@ -362,6 +365,14 @@ class Face final : public Geometry::FaceGeometry, public FaceData {
     /// The element owning this face, only valid if the face is owned by the
     /// current processor
     Element* getOwningElement() const;
+
+    void visitEntity(MeshEntityVisitor<>& visitor) final {
+        visitor.visit(*this);
+    }
+
+    void visitEntity(MeshEntityVisitor<true>& visitor) const final {
+        visitor.visit(*this);
+    }
 
    private:
     Element* elementLeft_;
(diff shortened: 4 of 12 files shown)


kernel/Base/Edge.cpp
```diff
@@ -111,7 +111,7 @@ bool Edge::isOwnedByCurrentProcessor() const {
     return elements_.size() > 0 && elements_[0]->isOwnedByCurrentProcessor();
 }
 
-Element* Edge::getOwningElement() const {
+const Element* Edge::getOwningElement() const {
```
JensWehner, line 114: nice

kernel/Base/Edge.h
```diff
@@ -79,15 +80,16 @@ class Edge {
         return numberOfConformingDOFOnTheEdge_[0];
     }
 
-    std::size_t getLocalNumberOfBasisFunctions(std::size_t unknown) const {
+    std::size_t getLocalNumberOfBasisFunctions(
+        std::size_t unknown) const final {
         // TODO: LC, numberOfConformingDOFOnTheNode_ might be smaller than
```
JensWehner, line 85: is this still a valid comment?

kernel/Base/Element.h
```diff
@@ -74,7 +76,9 @@ class PhysicalElement;
 
 // class is final as a reminder that the virtual default destructor should be
 // added once something inherits from this class
-class Element final : public Geometry::ElementGeometry, public ElementData {
+class Element final : public Geometry::ElementGeometry,
```
JensWehner, line 79: So I am not a fan of multiple inheritance the new baseclass you added is admittedly the best candidate of the 3.

kernel/Base/Element.h
```diff
@@ -525,9 +530,19 @@ class Element final : public Geometry::ElementGeometry, public ElementData {
...
+    bool isOwnedByCurrentProcessor() const final;
     std::size_t getOwner() const;
 
+    const Element* getOwningElement() const final { return this; }
+
+    void visitEntity(MeshEntityVisitor<>& visitor) final {
+        visitor.visit(*this);
+    }
+
+    void visitEntity(MeshEntityVisitor<true>& visitor) const final {
```
JensWehner, line 542: the `true` is probably the read/write?

kernel/Base/MeshEntity.h
```diff
@@ -0,0 +1,145 @@
...
+ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+#ifndef HPGEM_MESHENTITY_H
+#define HPGEM_MESHENTITY_H
+
+#include "functional"
```
JensWehner, line 41:
```suggestion
#include <functional>
```

kernel/Base/Face.h
```diff
@@ -354,14 +360,22 @@ class Face final : public Geometry::FaceGeometry, public FaceData {
...
     /// The element owning this face, only valid if the face is owned by the
     /// current processor
-    Element* getOwningElement() const;
+    const Element* getOwningElement() const final;
+
+    void visitEntity(MeshEntityVisitor<>& visitor) final {
+        visitor.visit(*this);
+    }
+
+    void visitEntity(MeshEntityVisitor<true>& visitor) const final {
```
JensWehner, line 376: ConstMeshEntityVisitor ?

kernel/Base/MeshEntity.h
```diff
@@ -0,0 +1,145 @@
...
+        typename std::conditional<constify, typename std::add_const<T>::type,
+                                  T>::type;
+
+    virtual void visit(Constified<Base::Element>& element) = 0;
+    virtual void visit(Constified<Base::Edge>& edge) = 0;
+    virtual void visit(Constified<Base::Face>& face) = 0;
+    virtual void visit(Constified<Base::Node>& node) = 0;
+};
+
+using ConstMeshEntityVisitor = MeshEntityVisitor<true>;
```
JensWehner, line 73: I actually would either typedef both cases or use C++17 where you do not need the `<>`anymore

kernel/Base/MeshEntity.h
```diff
@@ -0,0 +1,145 @@
...
+ * Visitor for the MeshEntity
+ * @tparam constify Whether the MeshEntities are const or not
+ */
+template <bool constify = false>
+class MeshEntityVisitor {
+   public:
+    virtual ~MeshEntityVisitor() = default;
+    /// Helper to add const
+    template <typename T>
+    using Constified =
```
JensWehner, line 63: nice, learned sth.

kernel/Base/MeshEntity.h
```diff
@@ -0,0 +1,145 @@
...
+
+using ConstMeshEntityVisitor = MeshEntityVisitor<true>;
+
+/**
+ * Entity of the Mesh, that is an Element, Face, Edge or Node.
+ */
+class MeshEntity {
+   protected:
+    // Deletion of MeshEntities should be governed by the actual instances.
+    ~MeshEntity() = default;
```
JensWehner, line 81:
```suggestion
    virtual ~MeshEntity() = default;
```

kernel/Base/MeshEntity.h
```diff
@@ -0,0 +1,145 @@
...
+};
+
+using ConstMeshEntityVisitor = MeshEntityVisitor<true>;
+
+/**
+ * Entity of the Mesh, that is an Element, Face, Edge or Node.
+ */
+class MeshEntity {
+   protected:
+    // Deletion of MeshEntities should be governed by the actual instances.
```
JensWehner, line 80: you might get slicing then if you use a pointer to base class. I might be wrong, but would not take the risk

kernel/Utilities/GlobalIndexing.cpp
```diff
@@ -58,6 +58,64 @@ GlobalIndexing::GlobalIndexing(Base::MeshManipulatorBase *mesh, Layout layout,
...
+    if (meshDimension > 1) {
+        for (const Base::Node *node : element->getNodesList()) {
+            visitor(node);
+        }
+    }
+}
+
+/**
+ * For an owned element, visit the element and the boundary parts it owns in the
+ * cononical order.
```
JensWehner, line 86:
```suggestion
 * canonical order.
```

kernel/Utilities/GlobalIndexing.cpp
```diff
@@ -226,39 +252,15 @@ void GlobalIndexing::constructUnblocked() {
...
-            for (Base::Node *node : element->getNodesList()) {
-                if (node->getElement(0) == element) {
-                    for (std::size_t unknown : includedUnknowns_) {
-                        offsets_[unknown].nodeOffsets_[node->getID()] = index;
-                        index += node->getLocalNumberOfBasisFunctions(unknown);
-                    }
+                    OffsetVisitor visitor(offsets_[unknown]);
+                    entity->visitEntity(visitor);
+                    *(visitor.entityOffset_) = index;
+                    index += entity->getLocalNumberOfBasisFunctions(unknown);
```
JensWehner, line 261: I do not understand what this is doing maybe make it a function?

kernel/Utilities/GlobalIndexing.h
```diff
@@ -44,6 +44,9 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "../Base/Face.h"
 #include "../Base/MeshManipulatorBase.h"
 #include "../Base/Node.h"
+
+#include "Base/MeshEntity.h"
```
JensWehner, line 48: weird why is the path different

kernel/Utilities/GlobalIndexing.h
```diff
@@ -133,115 +141,37 @@ class GlobalIndexing {
...
-        const Offsets& offset = offsets_[unknown];
-        logger.assert_debug(offset.includedInIndex_,
-                            "Unknown not included % in the index", unknown);
-        const auto basisStart = offset.nodeOffsets_.find(node->getID());
-        logger.assert_debug(basisStart != offsets_[unknown].nodeOffsets_.end(),
-                            "No indices known for node %:%", node->getID(),
-                            unknown);
-        return basisStart->second;
+        ConstOffsetVisitor visitor(offset);
+        entity->visitEntity(visitor);
```
JensWehner, line 155: I am not sure if the name `visitEntity` should be used here. Maybe `beVisited` or so?

kernel/Utilities/GlobalIndexing.h
```diff
@@ -493,6 +423,78 @@ class GlobalIndexing {
...
+        void visit(Base::Edge& edge) final {
+            entityOffset_ = &visitOffsets_.edgeOffsets_[edge.getID()];
+        }
+        void visit(Base::Node& node) final {
+            entityOffset_ = &visitOffsets_.nodeOffsets_[node.getID()];
+        }
+
+        Offsets& visitOffsets_;
+        /// Pointer to the entry, possibly newly initialized
+        int* entityOffset_;
```
JensWehner, line 451: why a pointer, does it have to be `nullptr` why not a copy? Just asking. :)

kernel/Utilities/GlobalIndexing.h
```diff
@@ -493,6 +423,78 @@ class GlobalIndexing {
...
+            entityOffset_ = &visitOffsets_.faceOffsets_[face.getID()];
+        }
+        void visit(Base::Edge& edge) final {
+            entityOffset_ = &visitOffsets_.edgeOffsets_[edge.getID()];
+        }
+        void visit(Base::Node& node) final {
+            entityOffset_ = &visitOffsets_.nodeOffsets_[node.getID()];
+        }
+
+        Offsets& visitOffsets_;
```
JensWehner, line 449: could we have a `const Reference`

kernel/Utilities/GlobalIndexing.h
JensWehner: Looking at the `ConstOffsetVisitor` I might not understand this class

JensWehner (commented):
It certainly did not make the architecture worse. I like the abstraction layer. I think some names could be clearer concerning the visitor. Is there a specific reason why you suspect a performance problem, is this bottleneck code?

kernel/Utilities/GlobalIndexing.cpp
```diff
@@ -58,6 +58,64 @@ GlobalIndexing::GlobalIndexing(Base::MeshManipulatorBase *mesh, Layout layout,
...
 
+/**
+ * Visit the element and all its boundary parts in the canonical order
+ *
+ * @param element The element to visit
+ * @param meshDimension The dimension of the mesh
+ * @param visitor The visitor to use
+ */
+void visitElementParts(const Base::Element *element, std::size_t meshDimension,
+                       std::function<void(const Base::MeshEntity *)> visitor) {
```
JensWehner, line 69: you could also just make this a templated function and get rid of the std::function.

JensWehner: if this is performance critical, you put in lambdas mostly anyway.

kernel/Base/Node.h
```diff
@@ -155,11 +159,19 @@ class Node {
...
-    Element *getOwningElement() const;
+    const Element *getOwningElement() const final;
+
+    void visitEntity(MeshEntityVisitor<> &visitor) final {
+        visitor.visit(*this);
+    }
+
+    void visitEntity(MeshEntityVisitor<true> &visitor) const final {
+        visitor.visit(*this);
+    }
```
JensWehner, line 174: I am not sure the performance advantage is worth the readability, probably has to be benchmarked.

kernel/Base/Edge.h
Laar: Looks like it already works for face/node, so I doubt it is still up to date.

kernel/Base/Element.h
Laar: The hierarchy here is kinda strange. Both ElementGeometry is just the geometric information about the element, and from that point of view should either be a field or just inlined. ElementData is similar, except that it has quite a bit of data that should, in my view, not be in Element at all.

The new MeshEntity would indeed be better as base class.

kernel/Base/Element.h
Laar: Const/non-const with respect to the entity, but I have now improved it so that you have `MeshEntityVisitor` as the non const one and `ConstMeshEntityVisitor` as the const one. Though at the cost of having the sligthly worse named `MeshEntityVisitorT` as base for them both.

kernel/Base/MeshEntity.h
Laar: Fixed

kernel/Base/Face.h
Laar: Done

kernel/Base/MeshEntity.h
Laar: Thanks to a stackoverflow

kernel/Base/MeshEntity.h
Laar: I was not doing this by advice of (among others) http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-dtor-virtual

kernel/Utilities/GlobalIndexing.cpp
Laar: Done

kernel/Utilities/GlobalIndexing.h
Laar: The others are historically weird, this is the more correct one. But I'll fix them all.

kernel/Utilities/GlobalIndexing.h
Laar: I have renamed it to `accept`,  which seems to be the standard in all examples for the visitor pattern that I found.

kernel/Utilities/GlobalIndexing.h
Laar: The intention is that it allows editing the offset. So this is either
 - nullptr before the visit
 - a pointer to an actual offset after the visit.
Note though that it can't be a reference, as that could only be initialized in the constructor, while we only have the information after visiting.

kernel/Utilities/GlobalIndexing.h
Laar: (see also previous) That would prevent editing the offset.

kernel/Base/Node.h
Laar: What would you then suggest for the readability? Because I think this boiler plate is needed for a visitor pattern.

kernel/Utilities/GlobalIndexing.cpp
Laar: Yes that is a good idea.

kernel/Base/Element.h
JensWehner: Good choice

kernel/Base/MeshEntity.h
JensWehner: I see. Actually not a bad idea as long as you have raw pointer running around

kernel/Utilities/GlobalIndexing.h
JensWehner: Okay

JensWehner: Are we on c++17 yet ? Std::optional

kernel/Utilities/GlobalIndexing.h
JensWehner: Yeah makes sense

kernel/Base/Node.h
JensWehner: Okay I would write visitor.visit(node) in the code but this is also fine

kernel/Utilities/GlobalIndexing.h
Laar: > Are we on c++17 yet ?

Not yet, though looking at the gcc version on the cluster I think we can switch over at some point

JensWehner (approved):
I think these are great changes, the multiple inheritance bugs me the most otherwise it is a step in the right direction. 

I have not checked for performance impact.

kernel/Base/Node.h
Laar: But what would node then refer to?

kernel/Utilities/GlobalIndexing.cpp
```diff
@@ -58,6 +58,74 @@ GlobalIndexing::GlobalIndexing(Base::MeshManipulatorBase *mesh, Layout layout,
...
 }
 
+/**
+ * Visit the element and all its boundary parts in the canonical order
+ *
+ * @param element The element to visit
+ * @param meshDimension The dimension of the mesh
+ * @param visitor The visiting function to use
+ * @tparam CV The type of the visitor, should be equivalent to
+ *  std::function<void(const MeshEntity*)>
```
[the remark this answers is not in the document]
Laar: Hmm, looking at the comment I should modify it so that it should be a type which has a `operator(const MeshEntity*)`, of which the result is discarded. But I don't know of a way to actually assert this.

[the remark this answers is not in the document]
JensWehner: Actually I think the compilation will fail if the `operator(const MeshEntity*)` does not exist

kernel/Utilities/GlobalIndexing.cpp
Laar: I've added several lines of documentation instead, as this is quite one-of and it does not really make a nice function (as it is hardly separable from the surrounding code).

Laar (commented):
Some smaller improvement

> the multiple inheritance bugs me the most
That has been bugging me for the last ~2 years. But in this case I would say that the simple removals (inlining, or  adding a member and forwarding the calls) both are a lot of work without any benefit. The better design would be to move the data stored in `ElementData` out of `Element`, but that is quite substantial work without any direct advantage.

Performance wise I have checked some computations, where I did not see any effect beyond the regular runtime variation. The CI runs seem to be of equivalent timing, so no large impact there either.

PR #130, opened 2021-07-15
Discover mesh dimension

Laar:
Add a basic reader for the mesh information. This is extracted from the actual reader in `MeshManipulator` so that other places can easily access this information as well.

The main external use case so far is to read the dimension of the mesh. Before this (like in DGMaxEigen) we needed to pass this as a command line parameter.


JensWehner:
I have to fix my notification settings. Sorry please feel free to ping me on teams as well, if it needs to be done quicker.

kernel/Base/MeshFileInformation.h
```diff
@@ -0,0 +1,107 @@
...
+
+namespace hpgem {
+namespace Base {
+
+/**
+ * Reader for the header information in the HPGEM mesh format.
+ *
+ * This will read the meta data of the mesh.
+ */
+class MeshFileInformation {
```
JensWehner, line 53: nitpick: turn it into a struct, then all is public and it is more obvious that it is mostly a bundle of data 
```suggestion
struct MeshFileInformation {
```

JensWehner (approved):
Yeah really nice PR. 👍

Laar:
> I have to fix my notification settings. Sorry please feel free to ping me on teams as well, if it needs to be done quicker.

I will in the future, but I did take some time off so it was not a big problem.

PR #131, opened 2021-07-24
Fixes out of date CI ubuntu repository information

Laar:
The ubuntu images on the github CI are refreshed approximately weekly.
In the mean time the ubuntu repository may release new packages and old
packages may be removed. Installing dependencies without updating the
repository information will thus occasionally lead to trying to install
an updated package. Doing so will fail the build.

See also https://github.com/actions/virtual-environments/issues/1757


JensWehner (approved):
Good catch

PR #132, opened 2021-07-24
Restructuring of the Preprocessor files

Laar:
Restructuring of the Preprocessor (primarily the mesh file).

The mesh in the Preprocessor is composed of 3 related classes:
 - `Mesh` : The whole mesh consisting out of smaller parts
 - `MeshEntity` : The definition of an entity in the Mesh of some dimension, which primarily points to the neighbouring Elements
 - `Element` : The definition of MeshEntities of the largest dimension, which contain all topological information about the connected MeshEntities.

These were grouped into a single header + template implementation file. While this was logical, this file became very big and unwieldy. Making it especially harder to view the different classes side by side.

This PR tries to fix this by splitting the mesh headers and restructuring the library a bit. Specifically:
 - Each of the three mesh components now has its own pair (definition & template implementation) header files.
 - These 6 and related files that are needed to define a mesh are grouped into a folder. The related files are
   - Definition of element shapes
   - idtypes
   - Predeclarations of the templates that are needed due to import cycles
 - The `MeshSource` -> `Mesh` have been moved to `MeshFactory.h`. Not doing this is MeshSource prevents importing all of the mesh in the mesh readers.
 -  Several support classes that are not Preprocessor related have been moved to a `util` folder
 - The preprocessor-lib is defined in a new cmakefile

The major disadvantage of the current structure is that `Mesh`, `MeshEntity` and `Element` template implementations need to import each other for the implementation of the template functions. Thus resulting in import cycles.

Reviewing notes:
 - The actual code has not changed, though some methods have been removed if they were marked as unused. So I would not look too much at the code, but mainly the imports.
 - I'm all ears for good suggestions around the import cycles.
 - Idea: It may be good to have a single header to import all the mesh parts for the rest of the library


Laar:
Thinking a bit more about it, I see an alternative that may be better for the cyclic dependencies.

Use one top level header, e.g. `Mesh.h` with for example as contents:
```
// Imports

// Predeclarations
template<std::size_t dim>
class Mesh;
...

#include "MeshImpl/MeshEntity.h"
#include "MeshImpl/Element.h"
#include "MeshImpl/Mesh.h"

#include "MeshImpl/MeshEntity_Impl.h"
#include "MeshImpl/Element_Impl.h"
#include "MeshImpl/Mesh_Impl.h"
```
And in each of the headers in MeshImpl only import the relevant headers (things like vector), but not any of the other MeshImpl parts.

JensWehner:
sounds good.

Laar:
@JensWehner Done that, so now ready for review.

I would not read the code in depth, as that is just copy-pasted from the old `mesh.h` and `mesh_impl.h` headers. I would primarily look at the structuring and probably at the few lines of documentation at the top.

JensWehner (approved):
Good changes, it makes the whole code more readable. I would merge it as it is, because it is a substantial improvement.

PR #133, opened 2021-08-01
Preprocessor merge entities

Laar:
This PR introduces the possibility to merge lower dimensional MeshEntities (i.e. Nodes, Edges, Faces but not Elements). This is a way to add periodicity to the mesh after construction.

The motivation for adding this is that the current way of adding periodicity caused invalid meshes. Specifically, when having periodic boundary condition for a 2-fold rotation axis (180 degree rotation). The resulting mesh would have faces with 4 adjacent elements, which is impossible. An example of the same effect is explained in `MeshSource.h`.

The solution to this problem is to first construct the mesh without periodicity. All the required faces/edges are then available. Then by merging the nodes/edges/faces only on the periodic boundary we prevent the incorrect merging the internal MeshEntities.

As this PR is quite substantial some reviewing order advice:
 1. Look at the example in `MeshSource.h` for a description of the problem
 2. Look at `Element`/`MeshEntity`. These were updated with several changes
     * The functions like `addEntity`/`setEntity` were modified to keep the bidirectional links correct, instead of having to both modify the Element and MeshEntity
     * `MeshEntity` got the possibility to merge with a different `MeshEntity` leaving the second one unused (i.e. not connected)
     * As result of having unused `MeshEntities` we need to remove these, resulting in some renumbering
  3. Optionally look at `Mesh`, this is the place where the renumbering happens
  4. Look at `MergePlan` and its corresponding unit test, the latter may be useful as example of its use
  5. The rest
Note: The hpgem meshes were regenerated and the result was checked, see 7374637 for a description.

In addition to the tests here I have also tested it on an actual mesh with two  2-fold rotation axes. Where it eliminated the right number of faces.


applications/Preprocessor/lib/mesh/MeshImpl/MeshEntity.h
```diff
@@ -173,7 +204,26 @@ class MeshEntity {
...
+    // Actual implementation of merge depends on the dimension of the
+    // MeshEntity, with the zero dimensional one being slightly different due to
+    // having to deal with coordinates. Partial template specialization is not
+    // possible with member functions, this pair of functions is a poor man's
+    // substitute preventing having to specialize the whole class for just this
+    // function.
+    template <std::size_t d>
+    std::enable_if_t<d == entityDimension> merge(
+        MeshEntity<entityDimension, meshDimension>& source, tag<d>);
+    void merge(MeshEntity<entityDimension, meshDimension>& source, tag<0>);
```
Laar, line 226: @JensWehner Note this is a bit of a hack, if you have any good ideas please let me know.

applications/Preprocessor/lib/MergePlan.h
```diff
@@ -0,0 +1,474 @@
...
+    // representative MeshEntity. This map takes that representative id and maps
+    // it to the corresponding index of the EntityMergeGroup in 'groups'.
+    // Using an index allows resizing merges without invalidating the
+    // 'references'.
+    std::map<EntityGId, std::size_t> mergeGroup;
+    for (const std::size_t rawEntityId : groups) {
+        EntityGId entityId = EntityGId(rawEntityId);
+        EntityGId represantative = EntityGId(groups.findSet(rawEntityId));
+
+        auto iter = mergeGroup.find(represantative);
```
JensWehner, line 252:
```suggestion
        auto iter = mergeGroup.find(representative);
```

applications/Preprocessor/lib/MergePlan.h
```diff
@@ -0,0 +1,474 @@
...
+    std::vector<EntityMergeGroup> merges;
+    // Each merge group in 'groups' is associated with the id of single
+    // representative MeshEntity. This map takes that representative id and maps
+    // it to the corresponding index of the EntityMergeGroup in 'groups'.
+    // Using an index allows resizing merges without invalidating the
+    // 'references'.
+    std::map<EntityGId, std::size_t> mergeGroup;
+    for (const std::size_t rawEntityId : groups) {
+        EntityGId entityId = EntityGId(rawEntityId);
+        EntityGId represantative = EntityGId(groups.findSet(rawEntityId));
```
JensWehner, line 250:
```suggestion
        EntityGId representative = EntityGId(groups.findSet(rawEntityId));
```

applications/Preprocessor/lib/MergePlan.h
```diff
@@ -0,0 +1,474 @@
...
+        auto iter = mergeGroup.find(represantative);
+        if (iter != mergeGroup.end()) {
+            // Additional entity to add to the merge group
+            merges[iter->second].push_back(entityId);
+        } else {
+            // First entity of the EntityMergeGroup
+            std::size_t index = merges.size();
+            // Add singleton element as start of the group
+            merges.push_back({entityId});
+            mergeGroup[represantative] = index;
```
JensWehner, line 261:
```suggestion
            mergeGroup[representative] = index;
```

applications/Preprocessor/lib/MergePlan.h
```diff
@@ -0,0 +1,474 @@
...
+    //     Map the coordinates to the corresponding right coordinates and store
+    //     the set of right coordinates with a reference to the left-entity.
+    //  2. Find all eligible right entities, and check if there is a matching
+    //     left-entity.
+
+    // Eligible left-entities, indexed by the mapped coordinates (i.e.
+    // coordinates in the right set).
+    std::map<std::set<CoordId>, EntityGId> leftCandidates;
+
+    for (const EntityGId& leftElementId : elements.leftElements) {
```
JensWehner, line 340: I find this hard to read and understand, maybe more functions help?

applications/Preprocessor/lib/MeshSource.h
```diff
@@ -190,6 +191,36 @@ class MeshSource2 {
...
+     * in each operation the coordinates identified by the indices on 'left' of
+     * the map are merged with the coordinates identified by the indices on the
+     * 'right' of the map. Each of these maps is fed to
+     * MergePlan#computeMergePlan for the actual merging.
+     *
+     * Note that unlike identifying coordinates beforehand (using Coord#nodeId),
+     * doing a merge via this route will prevent several problems in mesh
+     * construction. For example take two triangles in the configuration:
+     *
+     *  1 -- 0 -- 2
```
JensWehner, line 208: where is the pbc?

applications/Preprocessor/lib/centaur.cpp
```diff
@@ -325,15 +326,16 @@ void CentaurReader::readPeriodicNodeConnections() {
             // -1 due for Fortran indices
             std::size_t node0 = pairs[2 * i + 0] - 1;
             std::size_t node1 = pairs[2 * i + 1] - 1;
-            // Merge the nodes
-            connectedCoords.unionSets(node0, node1);
+            if (node1 == std::numeric_limits<std::uint32_t>::max()) {
```
JensWehner, line 329: urgh that is annoying

applications/Preprocessor/lib/mesh/MeshImpl/Mesh.h
```diff
@@ -184,7 +185,7 @@ class Mesh {
         dimension>>&
         getEntities() const;
     template <int entityDimension>
-    MeshEntity<entityDimension, dimension> getEntity(EntityGId i) const {
+    MeshEntity<entityDimension, dimension>& getEntity(EntityGId i) {
```
JensWehner, line 188: maybe also want a `const`version

applications/Preprocessor/lib/mesh/MeshImpl/Mesh.h
```diff
@@ -194,6 +195,30 @@ class Mesh {
...
+     * @return The number entities
+     */
+    std::size_t getNumberOfEntities(std::size_t entityDim) const {
+        if (entityDim == dimension) {
+            return elementsList.size();
+        } else {
+            logger.assert_debug(
+                entityDim < dimension,
+                "Asking for entities with dimension higher than the mesh.");
+            return getNumberOfEntities(entityDim, itag<dimension - 1>{});
```
JensWehner, line 218: I do not understand this.

applications/Preprocessor/lib/mesh/MeshImpl/Mesh.h
```diff
@@ -261,7 +292,14 @@ class Mesh {
     /// localIndex of the Element.
     template <std::size_t d>
     bool checkEntities(tag<d> dimTag) const {
```
JensWehner, line 294: Can we give this a more descriptive name?

applications/Preprocessor/lib/mesh/MeshImpl/MeshEntity.h
```diff
@@ -173,7 +204,26 @@ class MeshEntity {
...
     /// element.
-    void addElement(EntityGId elementID, EntityLId EntityLId);
+    void addElement(EntityGId elementID, EntityLId localEntityIndex);
+
+    /// Reverse of addElement, remove an Element that this MeshEntity is part
+    /// of.
+    ///
+    /// \param elementId The id of the element to remove
+    /// \param localEntityIndex The local index on the element of this
+    /// MeshEntity (before removal).
```
JensWehner, line 214: what happens at an id, after the element is removed?

applications/Preprocessor/lib/mesh/MeshImpl/MeshEntity.h
JensWehner: that is fine, limitation of the language.

applications/Preprocessor/lib/mesh/MeshImpl/MeshEntity_Impl.h
```diff
@@ -145,13 +145,89 @@ std::vector<EntityGId> MeshEntity<
...
+    while (!source.elementIDs.empty()) {
+        Element<meshDimension>& element = source.getElement(0);
+        EntityLId localId = source.getLocalIndex(0);
+        if (entityDimension != 0) {
+            element.template setEntity<entityDimension>(localId,
+                                                        getGlobalIndex());
+        }
+    }
+}
+
```
JensWehner, line 173: hmmm... maybe use a private template function to remove most of the duplicate code?

applications/Preprocessor/lib/mesh/MeshImpl/Mesh_Impl.h
```diff
@@ -234,8 +229,62 @@ void Mesh<dimension>::fixEntity(Element<dimension>& element,
...
-        element.getGlobalIndex(), index);
+}
+
+template <std::size_t dimension>
+template <int d>
+void Mesh<dimension>::removeUnusedEntities(itag<d> dimTag) {
+    std::vector<EntityGId> renumbering;
+    // Fill renumbering with invalid data
+    std::size_t currentCount = this->meshEntities[dimTag].size();
+    renumbering.resize(currentCount, EntityGId(-1));
```
JensWehner, line 240: Can we do it in the constructor of the std::vector?

applications/Preprocessor/lib/mesh/MeshImpl/Mesh_Impl.h
```diff
@@ -234,8 +229,62 @@ void Mesh<dimension>::fixEntity(Element<dimension>& element,
...
+    EntityGId newId = EntityGId(0);
+    // Whether we are actually removing entities
+    bool removing = false;
```
JensWehner, line 243: maybe `removed` is a better name.

JensWehner (changes requested):
Interesting change, I still do not fully get the whole structure. I will review again.

applications/Preprocessor/lib/MergePlan.h
Laar: I've split of the left into a function. The whole loop over elements and then over boundary MeshEntities has also been abstracted away. So hopefully that makes things a bit easier.

applications/Preprocessor/lib/MeshSource.h
Laar: Added a documenting note this is for a rotation periodicity around node 0.

applications/Preprocessor/lib/MeshSource.h
```diff
@@ -190,6 +191,36 @@ class MeshSource2 {
...
+     *     \ | /
+     *       3
+     * Identifying coordinates 1 & 2 during construction will merge edges:
+     *  - 0-1 with 0-2 and
+     *  - 1-3 with 2-3
+     *  The second one clearly being unintended. By performing the merge {0->0,
+     * 1->2} the coordinate 3 is not included. Hence, only the edges 0-1 and 0-2
+     * will be collapsed to a single edge, leaving edges 1-3 and 2-3 untouched.
+     */
+    virtual const std::vector<std::map<std::size_t, std::size_t>>& getMerges() {
```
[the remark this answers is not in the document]
Laar: As backwards compatibility and to simplify writing future sources without any merges. Though could be inlined as well.

applications/Preprocessor/lib/centaur.cpp
Laar: Yes indeed, was quite frustrated by it :(.

applications/Preprocessor/lib/mesh/MeshImpl/Mesh.h
Laar: Most of this I treat as a problem when I need it, as I'm not sure about the pass by value/reference of most of these functions.

applications/Preprocessor/lib/mesh/MeshImpl/Mesh.h
Laar: I've extracted the algorithm for this dynamic access into `TemplateArray`. I hope it is more clear in that context.

applications/Preprocessor/lib/mesh/MeshImpl/Mesh.h
Laar: Not sure, as it performs two checks:

 1. Check that the id of the entity corresponds to its index in the list of entities
 2. Check that if the MeshEntity has (element, localIndex) as bounding element, that the element has this entity at localIndex.

So I'm not really sure on how to provide a better name. Maybe if we split these checks into two functions.

applications/Preprocessor/lib/mesh/MeshImpl/MeshEntity.h
Laar: I've updated the documentation to make clear that this should only be used by `Element` to keep the link `MeshEntity -> Element` in sync with the link `Element -> MeshEntity` (and similar for addElement).

So it is the problem of the Element what happens at that specific index. The specific case where this is used is to overwrite the MeshEntity it links to. So it removes the link to one MeshEntity, calling this function to keep the reverse link up to date, ets a new MeshEntity and uses `MeshEntity#addElement` to set its link.

applications/Preprocessor/lib/mesh/MeshImpl/Mesh_Impl.h
Laar: Yes

applications/Preprocessor/lib/mesh/MeshImpl/Mesh_Impl.h
Laar: gone with `hasRemoved`

applications/Preprocessor/lib/mesh/MeshImpl/Mesh.h
JensWehner: Sounds like a good idea.

applications/Preprocessor/lib/mesh/MeshImpl/MeshEntity.h
JensWehner: maybe `Relink` is a better name?

JensWehner: I see what you are saying, but it is not intuitive from the functionname

applications/Preprocessor/lib/MergePlan.h
```diff
@@ -0,0 +1,545 @@
...
+
+    // For merging higher dimensional MeshEntities we need to know the elements
+    // that are adjacent to at least one of the coordinates for both the left
+    // and right coordinates.
+    MergeElements elements(mesh, pairing);
+    logger(INFO, "Found %/% left/right elements adjacent to a merge coordinate",
+           elements.leftElements.size(), elements.rightElements.size());
+
+    // Recurse over the dimensions for the MeshEntities of higher dimension.
+    computeMergePlanLevels(mesh, pairing, elements, merges, tag<1>{});
```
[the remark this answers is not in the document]
Laar: That one goes up, (1, 2, ...d-1).

PR #134, opened 2021-08-04
Reset basis function count

Laar:
The basis function count was not reset correctly when switching between different basis functions. 

When changing from CG to DG basis functions on the same mesh (probably does not happen in practice), the number of basis functions on Nodes, Edges and Faces was incorrect.

### DIFF
--- a/tests/unit/Base/BasisFunctionChange_UnitTest.cpp
+++ b/tests/unit/Base/BasisFunctionChange_UnitTest.cpp
@@ -1,0 +1,113 @@
+/*
+ This file forms part of hpGEM. This package has been developed over a number of
+ years by various people at the University of Twente and a full list of
+ contributors can be found at http://hpgem.org/about-the-code/team
+
+ This code is distributed using BSD 3-Clause License. A copy of which can found
+ below.
+
+
+ Copyright (c) 2021, University of Twente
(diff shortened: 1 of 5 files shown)


tests/unit/Base/BasisFunctionChange_UnitTest.cpp
```diff
@@ -0,0 +1,113 @@
...
+// order of the quadrature).
+
+void initMPI() {
+#ifdef HPGEM_USE_MPI
+    int initialized;
+    MPI_Initialized(&initialized);
+    if (!initialized) {
+        MPI_Init(nullptr, nullptr);
+    }
+    std::atexit([]() { MPI_Finalize(); });
```
JensWehner, line 64: nice

JensWehner (approved):

JensWehner:
Good catch.

tests/unit/Base/BasisFunctionChange_UnitTest.cpp
Laar: ~~Inspiration~~ Copied from `CommandLineOptions`

Laar:
I'll reformat the code this evening at home, easier than on the laptop.

PR #135, opened 2021-08-11
Rotated face connections

Laar:
This PR resolves a problem when you have periodic boundary faces with some kind of rotation.

Each face can be observed from both the left and right element. For periodic boundary faces such a description would give different coordinates for the corners depending on whether one views it from the left or right element. With pure translation periodicity any direction vector stays the same independent on whether view from the left or right element. However, when there is also some rotation in the periodicity this is no longer the case. For example consider the normal direction, this will be rotated.

This rotation poses a problem for computing face integrals on such boundary faces. One needs to express all vectors with respect to either the left or right orientation. In this PR we choose the left side as standard, and thus need to transform the vectors from the right side.

The PR consists of several related parts:
 1. Changes to `CoordinateTransform` to allow more general sources of the Jacobian (we need this for the transformation right -> left face)
 2.  Adding the transformation matrix to `PhysicalFace`
 3. Update `PhysicalFace` to transform the right quantities. To keep caching the quantities a layer of caches is added.
 4. Helper for the lazy caching in `PhysicalFace`

TODO: 

- [x] Recheck `PhysicalFace` that all required quantities are transformed
- [x] Check coverage results for `PhysicalFace`


kernel/Base/CoordinateTransformationData.h
```diff
@@ -0,0 +1,115 @@
...
+ * function values based on a coordinate transformation.
+ *
+ * @tparam DIM The dimension of the coordinate change.
+ */
+template <std::size_t DIM>
+class CoordinateTransformationData {
+    // Design note: This class is used to transform function values. The usual
+    // application is to transform multiple functions in a row. For example
+    // transforming all the basis functions at a fixed quadrature point. Hence,
+    // the values of this function will be request multiple times.
```
JensWehner, line 60:
```suggestion
    // the values of this function will be requested multiple times.
```

kernel/Base/LazyCached.h
```diff
@@ -0,0 +1,94 @@
...
+#ifndef HPGEM_LAZYCACHED_H
+#define HPGEM_LAZYCACHED_H
+
+#include <functional>
+
+namespace hpgem {
+namespace Base {
+
+/**
+ * A cache for a single value that is computed lazily.
```
JensWehner, line 47: might be obvious but I would add that this is not thread safe.

kernel/Base/PhysicalElement.h
```diff
@@ -240,10 +246,10 @@ class PhysicalElement final {
     LinearAlgebra::MiddleSizeVector solutionDiv;
 
     Geometry::PointPhysical<DIM> pointPhysical;
-    Geometry::Jacobian<DIM, DIM> jacobian, transposeJacobian,
-        inverseTransposeJacobian;
+    mutable Geometry::Jacobian<DIM, DIM> jacobian, transposeJacobian;
+    Geometry::Jacobian<DIM, DIM> inverseTransposeJacobian;
     double jacobianAbsDet;
-    double jacobianDet;
+    mutable double jacobianDet;
```
JensWehner, line 252: I know you have a warning earlier, but that can get problematic if multiple threads use it.

kernel/Base/LazyCached.h
```diff
@@ -0,0 +1,94 @@
...
+class LazyCached {
+   public:
+    /// Create a cache backed by a computation to obtain the value.
+    ///
+    /// Note the computation is only run when the value is first obtained. This
+    /// should be taken into account for any side effects that it may cause.
+    ///
+    /// Note: For the computation, the value is initialized to the default.
+    /// \param computation The computation to update the value
+    explicit LazyCached(std::function<void(T&)> computation)
```
JensWehner, line 61: I think instead you could also template it on the function, instead of using std::function.

kernel/Base/PhysicalFace.h
```diff
@@ -362,6 +365,24 @@ class PhysicalFace final {
...
+    void computeBasisFunctionDiv(std::vector<double>& values,
+                                 std::size_t unknown);
+
+    /// Updates the transformation information between the left and right faces
+    void updateLeftRightTransform();
+    /// Compute a set of direction vectors on the left/right face from
+    ///
+    /// helper for #updateLeftRightTransform()
+    LinearAlgebra::SmallMatrix<DIM, DIM> computeDirectionVectors(
+        std::array<LinearAlgebra::SmallVector<DIM>, DIM> points);
```
JensWehner, line 384: this makes a copy, would a reference work?

kernel/Base/LazyCached.h
Laar: Maybe good indeed to add, though practically there is quite little parts that will be thread safe.

JensWehner: yes, I agree. In my experience somehow cached things attract people to use multiple threads, It is a nitpick. 👍

kernel/Base/PhysicalElement.h
Laar: Is that because people would expect const-members to be completely thread-safe?

Besides, PhysicalElement is already a class that I would not think makes much sense in sharing between threads.

kernel/Base/LazyCached.h
Laar: What benefit would that have? Because I think it would then be converted in the constructor instead of in the code calling the constructor.

kernel/Base/PhysicalElement.h
JensWehner: Yeah, then please go ahead.

kernel/Base/LazyCached.h
JensWehner: yeah, I thought get rid of the lambda completely and save the T instead for performance reasons, but as the thing is cached anyway your std::function is properly the better thing. So please go ahead. :)

Laar: Problem there is that the T could change after each reset. The hard thing about generalizing over std::function is that you don't really have types for say the lambda and std::bind, etc.

kernel/Base/LazyCached.h
```diff
@@ -0,0 +1,196 @@
...
+        : compute_(computation), values_(){};
+
+    /**
+     * Utility constructor to use a non static member function as computation.
+     * @tparam C The class
+     * @param instance The instance to call the member function on
+     * @param func The member function used to update the cache.
+     */
+    template <typename C>
+    LazyVectorCached(C* instance, void (C::*func)(T&, std::size_t))
```
JensWehner, line 146: This function signature exists because it allows you to calculate multiple values at once

kernel/Base/PhysicalFace.h
```diff
@@ -156,27 +162,22 @@ class PhysicalFace final {
...
-                                                             std::size_t i);
     const LinearAlgebra::SmallVector<DIM>& basisFunctionCurl(
-        Side side, std::size_t i, std::size_t unknown);
+        Side side, std::size_t i, std::size_t unknown = 0);
 
     /// divergence of basis function i at the current reference point; indexing
     /// functions in the right element after functions in the left element
-    const double& basisFunctionDiv(std::size_t i);
-    const double& basisFunctionDiv(std::size_t i, std::size_t unknown);
+    const double& basisFunctionDiv(std::size_t i, std::size_t unknown = 0);
```
JensWehner, line 175: I am not a fan of returning const double& instead return just double, but it does not matter anyway. Why the default argument for unknown?

kernel/Base/PhysicalFace.h
```diff
@@ -362,6 +363,24 @@ class PhysicalFace final {
...
    private:
+    void resetLazyCaches(std::size_t currentUnknowns);
+    void computeBasisFunctionDeriv(
+        std::vector<LinearAlgebra::SmallVector<DIM>>& values,
+        std::size_t unknown);
+    void computeBasisFunctionCurl(
+        std::vector<LinearAlgebra::SmallVector<DIM>>& values,
+        std::size_t unknown);
+    void computeBasisFunctionDiv(std::vector<double>& values,
+                                 std::size_t unknown);
```
JensWehner, line 374: We cannot use the return type of these functions to return values, because of allocations? Would make the signature nicer.

kernel/Base/PhysicalFace_Impl.h
```diff
@@ -74,7 +83,11 @@ inline double PhysicalFace<DIM>::basisFunction(Side side, std::size_t i) {
     }
     logger.assert_debug(isInternal_,
                         "cannot find the right element for a boundary face");
-    return right.basisFunction(i);
+    double value = right.basisFunction(i);
+    if (requiresTransformation) {
+        value = transform_[0]->transform(value, rightVectorTransform);
+    }
+    return value;
```
JensWehner, line 90: could that be a separate function?

kernel/Base/PhysicalFace_Impl.h
```diff
@@ -305,6 +291,9 @@ inline void PhysicalFace<DIM>::basisFunction(
     }
     logger.assert_debug(isInternal_, "basis function index out of bounds");
     right.basisFunction(i - nLeftBasisFunctions[unknown], result, unknown);
+    if (requiresTransformation) {
```
JensWehner, line 294: would be somehow nice to factor this check out a little bit.

JensWehner: but might not be easy.

kernel/FE/BaseBasisFunction.h
```diff
@@ -145,6 +145,11 @@ class BaseBasisFunction {
         return 0;
     }
 
+    virtual LinearAlgebra::SmallVector<1> evalCurl(
+        const Geometry::PointReference<1>& p) const {
+        return {evalDeriv0(p)};
```
JensWehner, line 150: is that not consistent with the other evalCurl below?

tests/unit/Base/LazyCached_UnitTest.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+#include "../catch.hpp"
+#include "Base/LazyCached.h"
+
+#include <array>
+
+using hpgem::Base::LazyCached;
+using hpgem::Base::LazyVectorCached;
+
+TEST_CASE("Basic test", "[LazyCached]") {
```
JensWehner, line 47: Should we add a test for the class/memberfunction constructor as well?

JensWehner (approved):
Good changes, I have some nitpicks and questions otherwise good work.

kernel/Base/LazyCached.h
Laar: It is so that you don't have to place something like `[this](T& value, std::size_t index){this->computeValues(values, index);` or use `std::bind` with the long placeholders at every point where you want to invoke the constructor.

I tried it, but kept on repeating that initialization, so abstracted into this function.

kernel/Base/PhysicalFace.h
Laar: The original hpgem started with no unknowns, then transitioned into multiple unknowns but all with the same basis functions to be used for the approximation. So that is where the `basisFunctionDiv(std::size_t i)` signature came from.

The signature with unknown was added when unknowns with different basis functions were introduced. The old function remained which defaulted to using the 0-th unknown, but with a check that the basis functions for all unknowns are the same.

By having the default unknown this keeps compatibility with old codes/the common use case, but reduces the code duplication. The only effective change is that there is no check anymore that all basis functions are the same.

kernel/Base/PhysicalFace.h
Laar: That is indeed the idea. This is called for every quadrature point on every face. Each triangular face has say 4 quadrature points and a mesh can easily have several thousands of faces. The only upside is that the vector usually stays the same size (that is why there is a split internal and boundary faces in `PhysicalFace`).

kernel/Base/PhysicalFace_Impl.h
Laar: I would not see the benefit, as each of the values (normal, derivative, curl and divergence) would need a different transformation.

The general structure is the same for half of them, but there is quite a bit of small variability. So I don't know a good way to capture that.

kernel/FE/BaseBasisFunction.h
Laar: What is not consistent, the signature or the implementation?

Because the signature matches that of evalCurl for 2/3D in BaseBasisFunction. The difference here is that these values are computed relatively infrequently.

tests/unit/Base/LazyCached_UnitTest.cpp
Laar: I've added a test for `LazyCached` not much, but at least demonstrates its correct working.

kernel/FE/BaseBasisFunction.h
JensWehner: I mean in the other dimensions you throw an assert on evalCurl.

kernel/Base/PhysicalFace_Impl.h
JensWehner: okay

kernel/Base/LazyCached.h
JensWehner: I was not clear enough, I expected a signature like

`
```suggestion
    LazyVectorCached(C* instance, void (C::*func)(T&),std::size_t size)
```

Laar: Oh ok, that clarifies it. Then no, it does not allow you to compute multiple values at once. It allows a 'vector' of LazyCached values that are backed by a parametrized computation: `vector<LazyCached<T>>`. So the computation gets both the entry that is updated and the index of that entry (which can then affect the computational result).

With the signature you suggested there are two differences:

 1. The size argument at the end, that could equally well be added to the current method. But at the moment it is not used, so also not build.
 2. The computation does not have the index of the entry in the vector that is being updated. The problem with this is that all the entries share the exact same computation. So there is no natural way to for entries at different position to be different.

The original idea of calculating multiple values at once, I would expect to be the type `LazyCached<vector<T>>`. Note the difference in order with `vector<LazyCached<T>>`, which is the equivalent of LazyVectorCached. In the former the vector as a whole is lazy, while for the latter the entries are lazy.

JensWehner: okay thanks for the explanation

kernel/FE/BaseBasisFunction.h
Laar: Hmm, may indeed be a better approach.

PR #136, opened 2021-08-13
Reduce circular dependencies

Laar:
Offshoot of #135 aiming to reduce the circular dependencies in the includes.

The primary work is on decoupling PhysicalElement/Face from CoordinateTransformation. In #135 this was done for the value transformations (e.g. transformCurl). This PR removes the integrand scaling factors from CoordinateTransformation and moves these to the respective integral classes. Thereby removing the need for PhysicalElement/Face in CoordinateTransformation.

Secondly, PhysicalElement was used in Element to evaluate the solution, but that method was only used in PhysicalElement. By inlining the implementation the dependency of Element on PhysicalElement is removed.

Some minor other fixes that were found during debugging are also included.

After this PR the circular dependencies in the kernel are reduced to pairs `TemplateClass.h` and `TemplateClass_Impl.h`.

@JensWehner I remember that you tried reodering the imports at some point and failed due to the circular dependencies. This might be a time to retry that.


Laar:
Additionally added during the debugging and fixing stage:
 - Remove the dependency on L2Norm from Geometry, making the directory-include-graph much better.

JensWehner (approved):
Good change nothing to criticize unfortunately 👍

PR #137, opened 2021-08-13
Remove l2norm after #136 

Laar:
In #136 `Point#l2Norm` was added to break the dependency of Geometry on `Base/L2Norm`.

This followup PR replaces all remaining uses of `Base/L2Norm`  by member functions `l2Norm` and `l2NormSquared`.

The changes are:
 - Adding the required methods to `MiddleSizeVector`
 - Inling each of the remain L2Norm's (SmallVector, Point, MiddleSizeVector)
 - Cleaning up some crappy inlines
 - Remove the now obsolete unit test.

Recommendation: First merge #136, as the diff of this PR will be much more straight forward afterwards.


JensWehner (approved):
Nice change

PR #138, opened 2021-08-19
Dgmax Source

Laar:
Starting work on a separate program that solves the Maxwell Source problem, similar to DGMaxEigen that solves the eigenvalue problem. Such a problem has some source term (e.g. an emitter or an outside plane wave) and solves for the resulting field in a domain.

The current program is still relatively minimal and inflexible. For examples, the source term and boundary conditions are hardcoded, just as the values of the stability terms. This is by intention and will be made more flexible where needed in future PRs. For now the goal is to get a basic program into master, with further PRs adding more functionality where needed.


applications/DG-Max/Programs/DGMaxSource.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+    Base::parse_options(argc, argv);
+    initDGMaxLogging();
+    printArguments(argc, argv);
+
+    auto start = std::chrono::high_resolution_clock::now();
+
+    Base::MeshFileInformation meshInfo =
+        Base::MeshFileInformation::readInformation(meshFileName.getValue());
+    const std::size_t dimension = meshInfo.dimension;
```
JensWehner, line 79: I would just use `meshInfo.dimension`

JensWehner (approved):

applications/DG-Max/Programs/DGMaxSource.cpp
Laar: Indeed and is removed.

PR #139, opened 2021-08-20
Refactor Problems in DGMax

Laar:
The problem definitions for DGMax define many functions that go into the
PDE, for example boundary values and source terms. These functions are
refactored here to be more readable. In particular the follow changes
have been made:
 - Replaced returning the results by reference to just by value, thus:
   refactoring  void f(...., Vector<DIM>& result) to Vector<DIM> f(....)
 - Removed passing both a PointPhysical and a PhysicalFace. The latter
   contains the current point of evaluation, removing the need for the
   former.

The rest of the refactoring is just adjusting the code to work with the new signatures. The only cleanup considered was removing now unused variables.

Note #138 introduces a new source problem, so depending on merge order one of them needs to be adjusted.


JensWehner (approved):
Yeah good change much more functional and reabable

PR #140, opened 2021-08-23
Refactor DGMaxDiscretization matrix construction

Laar:
The primary step of this refactoring is to take the complicated matrix assembly step and restructure it. The original  implementation had a loop over the elements/faces that did several things:
 - Run the integrator
 - Delegate the integrand to a function
 - Postprocess the resulting matrix/matrices for the mass matrix
 - Store the matrices as element/face matrix
 - [not modified] Compute element/face vectors

This refactoring unclutters this process to only two function calls:
 1. Call a function to compute the face/element matrices
 2. Call a function to post process the matrices

The first function does the integration for each type of matrix and stores them in the element/face. The second one takes the matrix and the and post processes them as required. I decided to keep them in this fashion, as this would give the posibility of optimizing the element/face integrals at some point.

### DIFF
--- a/applications/DG-Max/Algorithms/DGMaxDiscretization.h
+++ b/applications/DG-Max/Algorithms/DGMaxDiscretization.h
@@ -35,6 +35,8 @@ ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
+#ifndef HPGEM_APP_DGMAXDISCRETIZATION_H
+#define HPGEM_APP_DGMAXDISCRETIZATION_H
 
 #include <functional>
 #include <map>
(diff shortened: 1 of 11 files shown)


applications/DG-Max/Algorithms/DGMaxDiscretization.h
```diff
@@ -212,6 +209,12 @@ class DGMaxDiscretization : public DGMaxDiscretizationBase {
                               InputFunction exactValue) const;
 
     const bool includeProjector_;
+    MassMatrixHandling matrixHandling_;
+
+    std::vector<std::shared_ptr<Base::CoordinateTransformation<DIM>>>
```
JensWehner, line 214: why shared_ptr?

JensWehner (approved):
Yeah, that also makes sense. Good job!

applications/DG-Max/Algorithms/DGMaxDiscretization.h
Laar: Not really thought about it, because that is what several classes require.

But looking at it now it seems the logical choice. Several classes need it and there is no reason for duplicating the instances, as the classes don't carry any state. Effectively they are just mathematical formula's.

Though a cleaner design might be to just have some instances on the stack and then passing pointers to those. Though that would be a boring and not so useful rewrite in the kernel.

PR #143, opened 2021-08-29
Cleanup divdgmax discretization

Laar:
I got frustrated with DivDGMaxDiscretization being far from elegant. This PR improves several parts.

The following things have been tackled:
 1. Reduce the number of integrators and transforms by sharing one set of them
 2. Extract a dimensionless base class for the discretization
 3. Move the code creating the element matrices inside a function.
 4. Replace the ~5 face integrands by 2, reducing the code duplication
 5. Remove excessive caching, this was recaching what was in PhysicalFace.
 6. Replace the way how input functions are passed to create element/face vectors
 7. Some minor type improvements.

Especially the face integrals of DivDGMaxDiscretization were significantly modified. The diff does not really look legible, split view might improve it, or otherwise look at the resulting functions without the original.

### DIFF
--- a/applications/DG-Max/Algorithms/DivDGMaxDiscretization.cpp
+++ b/applications/DG-Max/Algorithms/DivDGMaxDiscretization.cpp
@@ -47,6 +47,13 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 using namespace hpgem;
 
+// Definition of the constants to reference to.
+const std::size_t DivDGMaxDiscretizationBase::ELEMENT_MASS_MATRIX_ID;
+const std::size_t DivDGMaxDiscretizationBase::ELEMENT_STIFFNESS_MATRIX_ID;
+const std::size_t DivDGMaxDiscretizationBase::ELEMENT_SOURCE_VECTOR_ID;
+const std::size_t DivDGMaxDiscretizationBase::FACE_STIFFNESS_MATRIX_ID;
+const std::size_t DivDGMaxDiscretizationBase::FACE_BOUNDARY_VECTOR_ID;
+
 // Utility function
 struct FaceDoFInfo {
     std::size_t leftUDoFs;
@@ -1261,12 +1268,11 @@ double DivDGMaxDiscretization<DIM>::elementErrorIntegrand(
     return error.l2NormSquared();
 }
 
-template <std::size_t DIM>
-char fluxName(typename DivDGMaxDiscretization<DIM>::FluxType f) {
+char fluxName(typename DivDGMaxDiscretizationBase::FluxType f) {
     switch (f) {
-        case DivDGMaxDiscretization<DIM>::FluxType::BREZZI:
+        case DivDGMaxDiscretizationBase::FluxType::BREZZI:
             return 'b';
-        case DivDGMaxDiscretization<DIM>::FluxType::IP:
+        case DivDGMaxDiscretizationBase::FluxType::IP:
             return 'i';
         default:
             logger.assert_always(false, "Unknown flux type.");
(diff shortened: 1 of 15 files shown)


applications/DG-Max/Algorithms/DivDGMaxDiscretization.cpp
```diff
@@ -96,167 +142,87 @@ void DivDGMaxDiscretization<DIM>::initializeBasisFunctions(
...
-            new Base::HCurlConformingTransformation<DIM>()),
-        0);
-    faceIntegral.setTransformation(
-        std::shared_ptr<Base::CoordinateTransformation<DIM>>(
-            new Base::H1ConformingTransformation<DIM>()),
-        1);
+    Utilities::FaceLocalIndexing indexing;
+    indexing.reinit({0, 1});
 
     auto end = mesh.faceColEnd();
```
JensWehner, line 192: could that not be in the for loop declaration

Laar: What would the benefit be? Because that would constantly update the included unknowns every step instead of only once.

applications/DG-Max/Algorithms/DivDGMaxEigenvalue.cpp
```diff
@@ -359,9 +359,8 @@ void DivDGMaxEigenvalue<DIM>::solve(
     PetscErrorCode error;
     DGMaxLogger(INFO, "Starting assembly");
     discretization.initializeBasisFunctions(mesh_, order_);
-    discretization.computeElementIntegrands(mesh_, false, nullptr, nullptr,
-                                            nullptr);
-    discretization.computeFaceIntegrals(mesh_, nullptr, stab_);
+    discretization.computeElementIntegrands(mesh_, false, {});
```
JensWehner, line 362: I am still a fan of naming bool function arguments, but it is up to your discretion

applications/DG-Max/Algorithms/DivDGMaxHarmonic.cpp
```diff
@@ -198,14 +204,13 @@ void DivDGMaxHarmonic<DIM>::writeTec(std::string fileName) const {
...
-    const typename DivDGMaxDiscretization<DIM>::InputFunction& exactSolution)
-    const {
+    const typename DivDGMaxDiscretization<DIM>::InputFunction& exactSolution) {
     return discretization_.computeL2Error(mesh_, 0, exactSolution);
 }
 
 template <std::size_t DIM>
 double DivDGMaxHarmonic<DIM>::computeL2Error(
-    const ExactHarmonicProblem<DIM>& problem) const {
+    const ExactHarmonicProblem<DIM>& problem) {
```
JensWehner, line 213: the `const` is gone because of caching?

kernel/Utilities/FaceLocalIndexing.h
```diff
@@ -102,6 +102,32 @@ class FaceLocalIndexing {
...
     }
 
+    /**
+     * For a given unknown, what are the indices in the face matrx.
+     * @param unknown The unkown
+     * @param mapping Mapping from the i-th DoF/basisFunction to the
+     * corresponding row/column in the face matrix.
+     */
+    void getDoFMapping(std::size_t unknown,
+                       std::vector<std::size_t>& mapping) const {
```
JensWehner, line 112:
```suggestion
    std::vector<std::size_t> getDoFMapping(std::size_t unknown) const {
```

would this work?

JensWehner (approved):
Good refactoring, makes the code more readable.

applications/DG-Max/Algorithms/DivDGMaxDiscretization.cpp
JensWehner: Sorry, I meant the `auto end = mesh.faceColEnd();` could be inside the header of the for-loop

Laar: It could, but in the past I found out that it was quite inefficient. Apparently the construction of `faceColEnd()` is quite expensive. I don't know the exact numbers (may be in a previous commit message), but it was at least a few percent maybe even more than 10%.

applications/DG-Max/Algorithms/DivDGMaxHarmonic.cpp
Laar: It is because of moving the `Integral` classes up. They use `PhysicalElement/Face` for caching the basis functions and indeed they aren't const.

kernel/Utilities/FaceLocalIndexing.h
Laar: Could, but the primary idea here is to allow reusing the same vector without constantly having to reallocate the space.

Though to be fair, a lot of allocation is already being done.

JensWehner: Yes, okay but due to the resize call it might have to reallocate anyway. So yes I see the point, but I think it is not a big optimisation. But it is a fair argument, so feel free.
If you keep it, I would add a comment.

applications/DG-Max/Algorithms/DivDGMaxDiscretization.cpp
JensWehner: Okay, I would maybe add a comment, otherwise that is sensible.

kernel/Utilities/FaceLocalIndexing.h
Laar: Yes indeed. In practice though it would grow to the size of the largest number of basis functions and then stay at that size.

applications/DG-Max/Algorithms/DivDGMaxDiscretization.cpp
Laar: Done

applications/DG-Max/Algorithms/DivDGMaxEigenvalue.cpp
Laar: Even better, it was always false. So I just removed it.

PR #145, opened 2021-10-11
Dgmax boundary conditions2

Laar:
Adds a different type of boundary condition to (Div)DGMax. The previous code only allowed specifying the transverse part of the electric field (Dirichlet-style boundary condition). This PR changes adds the option to specify the transverse part of the magnatic field (= curl of electric field), which is a Neumann-style boundary condition.

Review recommendation:
First review
 - `BoundaryConditionType` giving a description of the boundary conditions
 -  `HarmonicProblem` the class that specifies which boundary condition should be used

### DIFF
--- a/applications/DG-Max/Programs/DGMaxSource.cpp
+++ b/applications/DG-Max/Programs/DGMaxSource.cpp
@@ -108,7 +108,7 @@ template <std::size_t dim>
 class TestingProblem : public HarmonicProblem<dim> {
 
    public:
-    TestingProblem() : pface(false){};
+    TestingProblem(){};
 
     double omega() const final { return 4.0e-2; }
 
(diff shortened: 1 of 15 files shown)


applications/DG-Max/Programs/DGMaxSource.cpp
```diff
@@ -123,6 +126,27 @@ class TestingProblem : public HarmonicProblem<dim> {
...
+        double nx = std::abs(std::abs(normal[0]) - 1.0);
+        if (nx < 1e-8) {
+            return DGMax::BoundaryConditionType::DIRICHLET;
+        } else {
+            return DGMax::BoundaryConditionType::NEUMANN;
+        }
+    }
+
+   private:
+    mutable Base::PhysicalFace<dim> pface;
```
JensWehner, line 149: ?

kernel/Utilities/GlobalVector.cpp
```diff
@@ -459,6 +459,21 @@ void GlobalPetscVector::writeTimeIntegrationVector(
     VecDestroy(&localB);
     CHKERRV(ierr);
 }
+
+void GlobalPetscVector::writeMatlab(Vec vec, const std::string& fileName) {
```
JensWehner, line 463: nice

applications/DG-Max/Programs/DGMaxSource.cpp
Laar: Eliminated its need. Was much easier than I though (should have looked earlier).

PR #146, opened 2021-10-12
Complex smallvector

Laar:
Introduces the option to have complex valued vectors of fixed size (`SmallVectorC`).

This is quite a bit more messy than expected due to several complicating factors:
 1. There are several differences between `double` and `complex<double>` in the standard library, like '<' working only on doubles, and `std::conj` always returning a `complex<double>`.
 2. The specialization of CrossProduct for vectors of size 2 and 3 now becomes a partial specialization
 3. The MiddleSizeVector (and matrix for that matter), can be real or complex valued depending on what petsc uses, which makes conversion SmallVector <-> MiddleSizeVector complicated. Removing this ambiguity from MiddleSizeVector is very time intensive, as it is intertwined with the Mesh.

This PR tries to balance time investment against doing "the right thing". In the end it would probably be better to replace all the linear algebra by an external library, but that would require a major investment of time.

### DIFF
--- a/kernel/LinearAlgebra/SmallVector.h
+++ b/kernel/LinearAlgebra/SmallVector.h
@@ -68,12 +68,24 @@ void computeCrossProduct(const GSmallVector<nRows, T>&,
                          const GSmallVector<nRows, T>&,
                          GSmallVector<nRows, T>&);
 
-inline MiddleSizeVector::type convertToMiddleSize(std::complex<double> v);
-inline MiddleSizeVector::type convertToMiddleSize(double v);
-// As out argument to allow overloading
-inline void convertFromMiddleSize(MiddleSizeVector::type v, double& out);
-inline void convertFromMiddleSize(MiddleSizeVector::type v,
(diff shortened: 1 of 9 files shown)


kernel/LinearAlgebra/SmallVector.h
```diff
@@ -36,137 +36,190 @@
...
+
+inline double conj(double v) { return v; }
+inline std::complex<double> conj(std::complex<double> v) {
+    return std::conj(v);
+}
+
+inline std::size_t hash(double v) noexcept { return std::hash<double>()(v); }
+inline std::size_t hash(std::complex<double> v) noexcept {
+
+    return std::hash<double>()(v.real()) + 17 * std::hash<double>()(v.imag());
```
JensWehner, line 98: why 17?

Laar: Just any other prime number to mix the hash values, no further particular reason.

PR #153, opened 2021-11-03
Unify dgmax discretizations

Laar:
Unifies the way the two dgmax discretizations work, by introducing an abstract base class. This allows for having only one solver for the harmonic problem instead of two.


Laar (commented):
Looks good

PR #160, opened 2021-12-21
Dgmax rectilinear pml

Laar:
Implements a rectilinear PML as described in the book by Monk.


applications/DG-Max/Algorithms/DGMaxDiscretization.cpp
```diff
@@ -84,6 +84,9 @@ void DGMaxDiscretization<DIM>::initializeBasisFunctions(
                              unknowns);
         mesh.useNedelecDGBasisFunctions(order_);
     }
+    for (Base::Element* element : mesh.getElementsList()) {
```
Laar, line 87: TODO: Remove

Laar: Removed

PR #163, opened 2022-02-21
Dgmax field pattern

Laar:
Add an abstraction for an electric field pattern without specified frequency.


Laar (commented):
Looks good on read through

PR #165, opened 2022-04-11
Add option to control MPI log output in DGMax

Laar:
Repeating the (almost) identical log output from multiple MPI processes will only clutter the output. However, for debugging it may be useful to see the log output from all of them. This commit adds an option --mpiloglevel allowing switching which messages get outputted by all processes.


HannoSpreeuw:
Not sure I can review this properly, since I barely know hpgem, apart from building, running benchmarks and profiling.

PR #169, opened 2022-07-15
Preprocessor add translation periodicity

Laar:
Allow adding translation periodic connections in the Preprocessor. This to augment meshes where the meshing process does not allow adding these connections.


Laar:
Self review

PR #174, opened 2022-08-23
Dgmax face integrals mpi fix

Laar:
Fixes to get DGMaxDiscretization to work with MPI meshes. Partially undoing overzealos fix in #173.


Laar:
Reviewed

PR #175, opened 2022-08-23
Implement a Jacobi Davidson Solver for solving Maxwell Equation 

NicoRenaud:


Laar (commented):
A few small things, but I have to look at it quite a bit longer. Also you might want to run `make format` over the code to appease the formatter.

NicoRenaud:
Thanks for the comment @Laar I shouldn't have requested your review it is far from being ready

applications/DG-Max/Algorithms/DGMaxEigenvalue.cpp
```diff
@@ -162,8 +164,12 @@ class DGMaxEigenvalue<DIM>::SolverWorkspace {
     /// Shell matrix P*product_, where P is the projection operator
     Mat shell_;
 
-    // Solver
-    EPS solver_;
+    // SLEPC Solver
+    EPS eps_solver_;
```
Laar, line 168: Could you rename these to `epsSolver_` and `jdmaxSolver_` to follow the conventions?

applications/DG-Max/Algorithms/DGMaxEigenvalue.cpp
```diff
@@ -484,28 +506,51 @@ PetscErrorCode compareEigen(PetscScalar ar, PetscScalar ai, PetscScalar br,
...
-        EPSSetProblemType(solver_, config_.useHermitian_ ? EPS_HEP : EPS_NHEP);
-    CHKERRABORT(PETSC_COMM_WORLD, err);
-    err = EPSSetWhichEigenpairs(solver_, EPS_WHICH_USER);
-    CHKERRABORT(PETSC_COMM_WORLD, err);
-    err = EPSSetEigenvalueComparison(solver_, compareEigen,
-                                     &(this->targetFrequency_));
-    CHKERRABORT(PETSC_COMM_WORLD, err);
-    err = EPSSetTarget(solver_, targetFrequency_ * targetFrequency_);
-    CHKERRABORT(PETSC_COMM_WORLD, err);
+    MatType mtype;
```
Laar, line 510: Is this used?

tests/self/DGMax/DGmax_Vacuum2D_SelfTest.cpp
```diff
@@ -147,5 +169,19 @@ int main(int argc, char** argv) {
...
+        // Some spacing between the results for the two seperate tests
+        std::cout << "\n\n";
+
+        std::cout << "Raw frequency table 3" << std::endl;
+        result3.printFrequencyTable(linear);
+
+        result3.printResultCode(10);
+
+        std::cout << "Cleanup results" << std::endl;
+        result3.filterResults(
```
Laar, line 182: Are there? Because the lowest eigenvalue that it should find is `0.8^2 + 0.9^2`. I would expect all zero eigenvalue to be filtered out.

Laar (changes requested):
Initial review for everything except the solver itself.

Laar:
I apparently broke the formatter configuration. So I expect that the formatter needs a rerun after merging #177 , because I see several different bracketing styles.

kernel/EigenSolvers/JacobiDavidsonMaxwell.cpp
```diff
@@ -0,0 +1,889 @@
...
+        ierr = computeSmallEigenvalues(small_evals, small_evects);
+        CHKERRABORT(PETSC_COMM_WORLD, ierr);
+
+        // create q
+        ierr = VecDuplicate(this->residue_vect, &q);
+        CHKERRABORT(PETSC_COMM_WORLD, ierr);
+
+        found = PETSC_TRUE;
+        idx_evect = 0;
+        while (found) {
```
Laar, line 724: Stability of the small eigenvectors (and eigenvalues) between processes will become quite important here. If they are not stable enough one process may accept the eigenvalues, while others might not.

kernel/EigenSolvers/JacobiDavidsonMaxwell.cpp
```diff
@@ -0,0 +1,889 @@
...
+                ierr = BVInsertVec(Q, Q_current_size, q);
+                CHKERRABORT(PETSC_COMM_WORLD, ierr);
+                this->Q_current_size++;
+
+                if (this->Q_current_size >= nev) {
+                    stop = PETSC_TRUE;
+                    break;
+                }
+
+                // Compute [tmp_v1 .... tmp_vk-1] = V [s2, .... sk]
```
Laar, line 763: I think that quite a bit of this juggling with vectors might have been simplified by sharing a BV for [Q, V]. By using BVSetActiveColumns one can then select either the Q or V part. What this would do to the readability is a second question though.

kernel/EigenSolvers/JacobiDavidsonMaxwell.h
```diff
@@ -0,0 +1,144 @@
...
+ *
+ * Implementation note:
+ *
+ */
+
+namespace hpgem {
+
+namespace EigenSolvers {
+
+class JacobiDavidsonMaxwellSolver final {
```
Laar, line 74: A few notes on the documentation here:

It would be nice to give a short description of the eigenvalue problem that is solved. Something like Ax = lambda x, constraint to C x = 0

There are several variables that are rather cryptically named for someone not having the paper next to the code. It might be nice to describe the content of these:

 - Y = C^H, H = C^Y (needed both for removing the space spanned by C)
 -  V = Eigenvalue search space 
 -  Q/Qt = Converged eigenvalues (I have no clue so far what the difference is between these)
 -  search/res_vect intermediate vectors for the search direction and residual.
 - eta = shift in the correction equation
 - tau = eigenvalue target?

kernel/EigenSolvers/JacobiDavidsonMaxwell.h
```diff
@@ -0,0 +1,144 @@
...
+    PetscInt search_space_current_size = 0;
+    PetscInt V_current_size = 0;
+    PetscInt Q_current_size = 0;
+    PetscInt Qt_current_size = 0;
+    PetscInt nconverged = 0;
+    PetscReal tolerance = 1E-3;
+
+    Mat A, C;
+    Mat Y, H;
+    BV Qt, Q, V;
```
Laar, line 131: As noted above, I've got no clue what the difference is between Qt and Q. Could you elaborate?

Laar (changes requested):
Looked at the solver itself. Looks like a good implementation with a few places with some problems. 

Also added a lot of places where you might have done some things slightly more efficiently by reordering some of the operation. Most of these don't need to be adjusted.

NicoRenaud:
@Laar  I've made all the changes you suggested. The formatter fails for some reason that I don't quite get ... 
We should also may change the exact requirement for clang-format-10 as it's not available on ubuntu 22 anymore.

applications/DG-Max/Algorithms/DGMaxEigenvalue.cpp
```diff
@@ -578,21 +621,28 @@ void DGMaxEigenvalue<DIM>::SolverWorkspace::initStiffnessMatrixShifts() {
...
 void DGMaxEigenvalue<DIM>::SolverWorkspace::extractEigenVectors() {
     std::swap(eigenpairs_, previousEigenpairs_);
-    eigenpairs_.loadEigenpairs(solver_, tempFieldVector_);
+    if (config_.use_jdmax_) {
+        eigenpairs_.loadEigenpairs(jdmaxSolver_, tempFieldVector_);
+    } else {
+        eigenpairs_.loadEigenpairs(epsSolver_, tempFieldVector_);
+    }
+      
     // Reorder
```
Laar, line 630: Is this code still needed?

applications/DG-Max/Programs/DGMaxEigen.cpp
```diff
@@ -254,8 +282,9 @@ class DGMaxEigenDriver : public AbstractEigenvalueSolverDriver<DIM> {
             // (or similarly predefined length). For the computation we assume
             // c=1, and assume the length scale a matches that of the mesh. Thus
             // the distance between x=0 and x=1 is assumed to be 'a'.
+            std::sprintf(freq_val, "%1.6f", frequency / (2 * M_PI) * lengthScale.getValue());
```
Laar, line 285: I'm wondering why you switch here to the old c-style of printing. If you just want to set the precision then `<< std::setprecision(6)` would do the trick as well.

kernel/Utilities/Eigenpairs.cpp
```diff
@@ -113,6 +114,35 @@ void Eigenpairs::loadEigenpairs(EPS eps, Vec sample) {
...
+    // Load the eigenpairs
+    for (PetscInt i = 0; i < converged; ++i) {
+        err = jdmax.getEigenPair(i, eigenvalues_[i], eigenvectors_[i]);
+        CHKERRABORT(PETSC_COMM_WORLD, err);
+    }
+    
+    // // Reset the ordering
+    std::iota(ordering_.begin(), ordering_.end(), 0);
+
+    // auto lcomp = [this](int a, int b) {
```
Laar, line 138: Is this still needed?

kernel/EigenSolvers/JacobiDavidsonMaxwell.h
```diff
@@ -0,0 +1,180 @@
...
+#include <math.h>
+#include <cstdlib>
+#include <vector>
+
+#include <chrono>  // For timing
+
+#include <utility>
+#include <valarray>
+
+/**
```
Laar, line 58: Duplicate partial comment

kernel/EigenSolvers/JacobiDavidsonMaxwell.cpp
```diff
@@ -0,0 +1,952 @@
...
+
+PetscReal JacobiDavidsonMaxwellSolver::computeSmallResidue(
+    const Mat &A, const Vec &x, const PetscScalar lambda)
+{
+    PetscReal resVal;
+    Vec resVec;
+    Vec tmp;
+
+    VecDuplicate(x, &resVec);
+    VecDuplicate(x, &tmp);
```
Laar, line 497: These two vectors don't seem to be destroyed

Laar (changes requested):
Looks quite improved, a few small comments still.

The issue with clang-format is probably best resolved by updating it to a newer version. But that probably should go with a reformatting of the code.

Laar (changes requested):
Getting close to finishing

PR #178, opened 2022-09-27
Output higher order mesh input in DGMaxEigen

Laar:
The written mesh by DGMaxEigen was using a linear mesh, irrespective of whether the mesh used curvilinear elements. This commit changes it to determine the geometric order of the mesh and use that to output the mesh sample.


Laar:
Probably good to check DGMaxSource as well.

PR #182, opened 2022-10-18
Add mpi tests

apalha:
I have added the draft MPI tests. For now I only implemented the parallel tests for [DGMax_Vacuum2D_SelfTest.cpp](https://github.com/hpgem/hpgem/compare/add_MPI_tests?expand=1#diff-2ee5068fd66ab224118514cac470288476401fdcaa2593d53530148fdb3b047b).

The main changes are:

1. [reGenerateMeshFiles](https://github.com/hpgem/hpgem/compare/add_MPI_tests?expand=1#diff-0ab83ce7a7cb6c47dde5b3b09bd05b9a99f3236f877ece966d366cbb55cc1a77)
Adds the generation of meshes with two partitions using -n 2 switch. Stored in folder `parallel_meshes`, so that it does not interfere with current workflow.
2. [tests/self/DGMax/testMeshes.h](https://github.com/hpgem/hpgem/compare/add_MPI_tests?expand=1#diff-078b1044d8a34eeb86027777798c02bdd9200c4b81fdd31a2df68a570c541ca9)
Adds function to return mesh filenames for parallel meshes (from the `parallel_meshes` folder). Adds function that returns either parallel or serial mesh filenames depending on how the executable is running (in serial or parallel with `-n 2` switch). This last function was added not to break backwards compatibility and for cases where only serial or parallel meshes are to be used and not both.
3. [...self/DGMax/DivDGmax_Vacuum2D_SelfTest.cpp → ...self/DGMax/DivDGMax_Vacuum2D_SelfTest.cpp](https://github.com/hpgem/hpgem/compare/add_MPI_tests?expand=1#diff-cdde86ba9727ff0d0390f25780be0e4b36d1e289b3dff4e66a73dac42db53c9f)
Replaces call to serial mesh filenames generation with general function that chooses serial and parallel mesh filenames. 
4. [conf/cmake/MasterMake.cmake](https://github.com/hpgem/hpgem/compare/add_MPI_tests?expand=1#diff-beccc88e5a57c88975efd8e9846d8626c42176cb0710147e9ec35208d9efcab0)
Adds generation of parallel tests with `-n 2 switch`. For now this is done only for `DivDGmax_Vacuum2D_SelfTest.cpp`. Can easily be extendend. These tests have the termination `_parallel`. This means this tests can be excluded if used `ctest -E parallel`, for example. These parallel tests are only added if hpgem is compiled with MPI and with METIS, i.e., `hpGEM_USE_MPI` and `hpGEM_USE_METIS` both set to `ON`.

As mentioned above, for now only DivDGmax_Vacuum2D_SelfTest.cpp is tested in parallel. If this general structure is accepted, the other tests can be extended also.

One important point is that the solver MUMPS is required for the parallel runs.

### DIFF
--- a/applications/DG-Max/Algorithms/DGMaxDiscretization.cpp
+++ b/applications/DG-Max/Algorithms/DGMaxDiscretization.cpp
@@ -805,12 +805,20 @@ std::map<typename DGMaxDiscretizationBase::NormType, double>
         for (typename Base::MeshManipulator<DIM>::FaceIterator it =
                  mesh.faceColBegin();
              it != end; ++it) {
+            if (!(*it)->isOwnedByCurrentProcessor()) {
+                continue;
+            }
             dgNorm += faceIntegrator_.integrate(
                 *it, [&](Base::PhysicalFace<DIM>& face) {
                     return faceErrorIntegrand(face, timeVector, electricField);
                 });
         }
     }
+#ifdef HPGEM_USE_MPI
+    MPI_Allreduce(MPI_IN_PLACE, &l2Norm, 1, MPI_DOUBLE, MPI_SUM, MPI_COMM_WORLD);
+    MPI_Allreduce(MPI_IN_PLACE, &hCurlNorm, 1, MPI_DOUBLE, MPI_SUM, MPI_COMM_WORLD);
+    MPI_Allreduce(MPI_IN_PLACE, &dgNorm, 1, MPI_DOUBLE, MPI_SUM, MPI_COMM_WORLD);
+#endif
 
     std::map<NormType, double> result;
     if (l2Wanted) result[L2] = sqrt(l2Norm);
--- a/applications/DG-Max/Algorithms/DivDGMaxDiscretization.cpp
+++ b/applications/DG-Max/Algorithms/DivDGMaxDiscretization.cpp
@@ -1278,6 +1278,9 @@ double DivDGMaxDiscretization<DIM>::computeL2Error(
                                              electricField);
             });
     }
+#ifdef HPGEM_USE_MPI
+    MPI_Allreduce(MPI_IN_PLACE, &error, 1, MPI_DOUBLE, MPI_SUM, MPI_COMM_WORLD);
+#endif
     return std::sqrt(error);
 }
 
--- a/applications/DG-Max/Algorithms/HarmonicSolver.cpp
+++ b/applications/DG-Max/Algorithms/HarmonicSolver.cpp
@@ -212,7 +212,7 @@ HarmonicSolver<DIM>::Workspace::Workspace(
     // Initialize the basis functions and indices after creating the matrices,
     // so that the matrices are not yet assembled.
     discretization_->initializeBasisFunctions(*mesh_);
-    indexing_.reset(mesh_, Utilities::GlobalIndexing::BLOCKED_GLOBAL);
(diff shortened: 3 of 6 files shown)


tests/self/DGMax/testMeshes.h
```diff
@@ -43,6 +43,35 @@
...
+                                    // (serial) or two (parallel) processors
+    MPI_Comm_size(MPI_COMM_WORLD, &numberOfProcessors);
+    logger.assert_always((numberOfProcessors > 0) &&
+                             (numberOfProcessors <= maxNumberOfProcessors),
+                         "Parallel run only up to two processors");
+    std::cout << "\nNumber of processors: " << numberOfProcessors;
+    switch (numberOfProcessors) {
+        case 1:
+            // Only one processor so we are running in serial mode
+            std::cout << "\nSerial mode!\n";
```
Laar, line 61:
```suggestion
            logger(INFO, "Serial mode");
```

tests/self/DGMax/testMeshes.h
```diff
@@ -43,6 +43,35 @@
...
+    bool isParallelRun = false;
+#ifdef HPGEM_USE_MPI
+    int numberOfProcessors = 1;
+    int maxNumberOfProcessors = 2;  // the test is designed to run only for one
+                                    // (serial) or two (parallel) processors
+    MPI_Comm_size(MPI_COMM_WORLD, &numberOfProcessors);
+    logger.assert_always((numberOfProcessors > 0) &&
+                             (numberOfProcessors <= maxNumberOfProcessors),
+                         "Parallel run only up to two processors");
+    std::cout << "\nNumber of processors: " << numberOfProcessors;
```
Laar, line 57:
```suggestion
    logger(INFO, "Number of processors: %", numberOfProcessors);
```

tests/self/DGMax/testMeshes.h
```diff
@@ -43,6 +43,35 @@
...
+    switch (numberOfProcessors) {
+        case 1:
+            // Only one processor so we are running in serial mode
+            std::cout << "\nSerial mode!\n";
+            isParallelRun = false;
+            break;
+        case 2:
+            // Two Processors so we are running in parallel mode (we can only
+            // run with 1 or 2 processors)
+            std::cout << "\nParallel mode!\n";
```
Laar, line 67:
```suggestion
            logger(INFO, "Parallel mode!");
```

Laar (commented):
Excelent work, a few minor changes.

Laar:
Looks like a very nice update, and I think this is a good approach.

 - Quite of a few of the test meshes are enumerated in `tests/self/TestMeshes.h`, so that would allow relative easy extension to multiprocessor runs by doing the same path-switching approach.
 - I will need to fix the centaur reader (.hyb files) so that it accepts the sample meshes again. As it has some problem with the older mesh format (compared the current meshes that I use).
 - It might be useful to move the serial meshes to their own folder as well, though this might break quite a few tests.

Laar:
Problem with the centaur (hyb) meshes should be fixed by #184

apalha:
Hi @Laar , thanks for your input! Following up with your three points:

- That is what I had in mind.
- Great.
- That was the reason why I did not move them to their own folder (I agree it is cleaner to place them also in their own folder). I will see if I can easily change the tests to satisfy that change.

As for the point you raised in Teams (MUMPS and Ubuntu): I will have a look into that.

PR #185, opened 2022-11-18
Add mpi tests plane wave fixes

Laar:
Several fixes to make DGMax_PlaneWave_SelfTest work in a multi process setting.

### DIFF
--- a/applications/DG-Max/Algorithms/DGMaxDiscretization.cpp
+++ b/applications/DG-Max/Algorithms/DGMaxDiscretization.cpp
@@ -805,12 +805,20 @@ std::map<typename DGMaxDiscretizationBase::NormType, double>
         for (typename Base::MeshManipulator<DIM>::FaceIterator it =
                  mesh.faceColBegin();
              it != end; ++it) {
+            if (!(*it)->isOwnedByCurrentProcessor()) {
+                continue;
+            }
             dgNorm += faceIntegrator_.integrate(
                 *it, [&](Base::PhysicalFace<DIM>& face) {
                     return faceErrorIntegrand(face, timeVector, electricField);
                 });
         }
     }
+#ifdef HPGEM_USE_MPI
+    MPI_Allreduce(MPI_IN_PLACE, &l2Norm, 1, MPI_DOUBLE, MPI_SUM, MPI_COMM_WORLD);
+    MPI_Allreduce(MPI_IN_PLACE, &hCurlNorm, 1, MPI_DOUBLE, MPI_SUM, MPI_COMM_WORLD);
+    MPI_Allreduce(MPI_IN_PLACE, &dgNorm, 1, MPI_DOUBLE, MPI_SUM, MPI_COMM_WORLD);
+#endif
 
     std::map<NormType, double> result;
     if (l2Wanted) result[L2] = sqrt(l2Norm);
--- a/applications/DG-Max/Algorithms/HarmonicSolver.cpp
+++ b/applications/DG-Max/Algorithms/HarmonicSolver.cpp
@@ -212,7 +212,7 @@ HarmonicSolver<DIM>::Workspace::Workspace(
     // Initialize the basis functions and indices after creating the matrices,
     // so that the matrices are not yet assembled.
     discretization_->initializeBasisFunctions(*mesh_);
-    indexing_.reset(mesh_, Utilities::GlobalIndexing::BLOCKED_GLOBAL);
+    indexing_.reset(mesh_, Utilities::GlobalIndexing::BLOCKED_PROCESSOR);
     resultVector_.reinit();
 
     PetscErrorCode err;
--- a/tests/self/ConvergenceTest.h
+++ b/tests/self/ConvergenceTest.h
@@ -111,12 +111,11 @@ void runConvergenceTest(
 	    // std::cout << "relativeAccuracy: " << testSet.relativeAccuracy << "\n";
 	    // std::cout << "expectedError: " << testSet.expectedErrors[j][i] << "\n";
 	    logger(INFO, "error: %", error);
-	    logger(INFO, "ExpectedError: %", testSet.expectedErrors[j][i]);
-	    logger(INFO, "ExpectedAccuracy: %", testSet.expectedErrors[j][i]);
-	    
-	    error = testSet.expectedErrors[j][i];
+            if (i < testSet.expectedErrors[j].size()) {
+                logger(INFO, "ExpectedError: %", testSet.expectedErrors[j][i]);
+            }
+	    logger(INFO, "ExpectedAccuracy: %", testSet.relativeAccuracy);
 	    observedErrors[j].push_back(error);
-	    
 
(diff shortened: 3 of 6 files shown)


applications/DG-Max/Algorithms/DGMaxDiscretization.cpp
```diff
@@ -805,12 +805,20 @@ std::map<typename DGMaxDiscretizationBase::NormType, double>
         for (typename Base::MeshManipulator<DIM>::FaceIterator it =
                  mesh.faceColBegin();
              it != end; ++it) {
+            if (!(*it)->isOwnedByCurrentProcessor()) {
```
apalha, line 808: With this you only work on what is owned by the current processor. Before it was operating on everything.

applications/DG-Max/Algorithms/DGMaxDiscretization.cpp
```diff
@@ -805,12 +805,20 @@ std::map<typename DGMaxDiscretizationBase::NormType, double>
...
+                continue;
+            }
             dgNorm += faceIntegrator_.integrate(
                 *it, [&](Base::PhysicalFace<DIM>& face) {
                     return faceErrorIntegrand(face, timeVector, electricField);
                 });
         }
     }
+#ifdef HPGEM_USE_MPI
```
apalha, line 817: I see, here you add the reduce, before it was assumed it was one single processor. So each was computing its own error norm and not sharing.

applications/DG-Max/Algorithms/HarmonicSolver.cpp
```diff
@@ -212,7 +212,7 @@ HarmonicSolver<DIM>::Workspace::Workspace(
     // Initialize the basis functions and indices after creating the matrices,
     // so that the matrices are not yet assembled.
     discretization_->initializeBasisFunctions(*mesh_);
-    indexing_.reset(mesh_, Utilities::GlobalIndexing::BLOCKED_GLOBAL);
+    indexing_.reset(mesh_, Utilities::GlobalIndexing::BLOCKED_PROCESSOR);
```
apalha, line 215: Here you set the indices for each processor instead of using the global indexing.

tests/self/DGMax/DGMax_PlaneWave_SelfTest.cpp
```diff
@@ -120,7 +120,7 @@ class Driver : public AbstractHarmonicSolverDriver<2> {
         double expectedOutFlux = 0.0;
 
         for (Base::Face* face : result.getMesh().getFacesList()) {
-            if (face->isInternal()) {
+            if (face->isInternal() || !face->isOwnedByCurrentProcessor()) {
```
apalha, line 123: Same here for faces to compute fluxes.

tests/self/TestMeshes.h
```diff
@@ -158,7 +158,7 @@ std::vector<std::string> getUnitSquareTriangleMeshes(
         meshFilenames = getSingleProcessorUnitSquareTriangleMeshes();
     }
 
-    return meshFilenames;
+    return limit(meshFilenames, minLevel, maxLevel);
```
apalha, line 161: My bad here. Well spotted.

tests/self/ConvergenceTest.h
```diff
@@ -111,12 +111,11 @@ void runConvergenceTest(
 	    // std::cout << "relativeAccuracy: " << testSet.relativeAccuracy << "\n";
 	    // std::cout << "expectedError: " << testSet.expectedErrors[j][i] << "\n";
 	    logger(INFO, "error: %", error);
-	    logger(INFO, "ExpectedError: %", testSet.expectedErrors[j][i]);
-	    logger(INFO, "ExpectedAccuracy: %", testSet.expectedErrors[j][i]);
-	    
-	    error = testSet.expectedErrors[j][i];
+            if (i < testSet.expectedErrors[j].size()) {
```
apalha, line 114: This was just to not get stuck in the first test, to see if I could see the pattern. This is to be removed.

applications/DG-Max/Algorithms/DGMaxDiscretization.cpp
Laar: Indeed, for reference we are basically computing an integral over all the elements or faces of the mesh, computing the local error on each part.

applications/DG-Max/Algorithms/HarmonicSolver.cpp
Laar: There are solution coefficients (known as Degrees of Freedom) for each variable in the original problem and each mesh part (Element, Face, etc.). These options set in which order they should be enumerated, and BLOCKED_PROCESSOR, is not suitable to use with the requirements from regular PETSc matrices.

tests/self/DGMax/DGMax_PlaneWave_SelfTest.cpp
Laar: Indeed

tests/self/ConvergenceTest.h
Laar: Ok no problem, just had to add it because it started to go out of bounds while testing.

applications/DG-Max/Algorithms/HarmonicSolver.cpp
apalha: The FEM (DG and CG) part I am very familiar with, my unfamiliarity is more with hpgem and PETSc. What do you mean by regular PETSc matrices? matseqaij? BLOCKED_PROCESSOR does indexing of the dofs per processor. BLOCKED_GLOBAL does the indexing globally, is that it?

Laar: With the regular ones I mean the mpimaij like ones. But not the one where you create a block matrix out of individual smaller PETSc matrices.

So to reword the thing about indexing, MatSetValues uses the global row and column index to set or update values in the matrix. So we need to define a mapping from the basis function for element e, belonging to process p, and variable v (e.g. pressure, velocity) to the global row/column in the matrix. The difference between the three options is how to create a global ordering/index with these three subindices.

There are three options that I build at some point into hpgem, which order the basis functions lexicographically using the triples:
 - `SEQUENTIAL`: (p,e,v)
 - `BLOCKED_PROCESSOR`: (p,v,e)
 - `BLOCKED_GLOBAL`: (v,p,e)
So the `BLOCKED_GLOBAL` groups all basis functions of a single variable together, then orders them by process for that process orders them by element. As all the basis functions of a single variable are grouped together, the resulting global matrix would be a block matrix (per variable). However, this means that the indices of a single process are not continuous. This is not allowed by PETSc's MAT_MPIAIJ, which requires them to be sequential. So there `BLOCKED_PROCESSOR` comes in, this first orders them by process, so that all the rows/columns of a single process are consecutive, and within those rows creates the block structure of ordering by variable. Lastly, the `SEQUENTIAL` style does not create a larger block structure and for each process orders them by element.

